Issue #8, opened 2020-07-13
Adding the `bin/envsetup.sh` file and docker-compose for Development env


anduonghien:
Should we add the `bin/envsetup.sh` file similar the https://github.com/nimblehq/rails-templates/blob/develop/bin/envsetup.s, so the developer can easily bootstrap the development env

olivierobert:
As commented on the PR, I recommend to use a Makefile for the Elixir template.

PR #9, opened 2020-07-13
Adding envsetup to boostrap development env

anduonghien:
https://github.com/nimblehq/elixir-templates/issues/8

## What happened

Adding the **Makefile** and **docker-compose.dev.yml** to help new developer could bootstrap the development env easily
 
## Insight

Just using docker-compose.
 
## Proof Of Work

1/ Start docker container

image

2/ mix test

image

3/ mix ecto.create

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,4 @@
+.PHONY: env_setup
+
+env_setup: ## Setup development environment
+	docker-compose -f docker-compose.dev.yml up -d
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: env_setup
+.PHONY: docker_setup
 
-env_setup: ## Setup development environment
+docker_setup:
 	docker-compose -f docker-compose.dev.yml up -d
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,11 +1,11 @@
-version: '3.2'
+version: "3.2"
 
 services:
   db:
     image: postgres:12.3
-    container_name: elixir_template
+    container_name: Nimble
     environment:
       - POSTGRES_HOST_AUTH_METHOD=trust
-      - POSTGRES_DB=elixir_template_development
+      - POSTGRES_DB=Nimble_development
     ports:
       - "5432:5432"
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -3,9 +3,9 @@ version: "3.2"
 services:
   db:
     image: postgres:12.3
-    container_name: Nimble
+    container_name: nimble
     environment:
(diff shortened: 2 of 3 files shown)


anduonghien:
> Good to add Docker Compose.
> 
> As for bootrapping the app, since by default there is no `bin` folder (contrary to Rails), I think we should use a Makefile instead. It's more standard and would not introduce a non-typical directory.

I have changed it to Makefile in f3b89c6 @olivierobert 

image

docker-compose.dev.yml
```diff
@@ -0,0 +1,11 @@
+version: "3.2"
+
+services:
+  db:
+    image: postgres:12.3
+    container_name: Nimble
+    environment:
+      - POSTGRES_HOST_AUTH_METHOD=trust
+      - POSTGRES_DB=Nimble_development
```
Tolsee, line 9: 🤔 
```suggestion
      - POSTGRES_DB=nimble_dev
```

anduonghien: Sure, will change it @Tolsee

Makefile
```diff
@@ -0,0 +1,4 @@
+.PHONY: env_setup
+
+env_setup: ## Setup development environment
```
olivierobert, line 3: How about simply `setup`?

anduonghien: I prefer to keep it as `env_setup` as the `setup` is bigger than the `env`, maybe later we have another `setup step` so we could separate it to different `step`, and if we want to combine that into 1 setup step, Makefile allow us to do that

```
setup: env_setup another_setup

env_setup:
 - xxx

another_setup:
 - yyy
```

olivierobert: To me, it feels like inheriting from the Rails project. And actually there is no environment setup at all, just the db 😆 How about `docker_setup` then? or `db_setup`?

anduonghien: We can go with `docker_setup` @olivierobert

Tolsee:
> What happened
Adding the bin/envsetup.sh and docker-compose.dev.yml to help new developer could boostrap the development env easily

Please update the PR description

docker-compose.dev.yml
```diff
@@ -0,0 +1,11 @@
+version: "3.2"
+
+services:
+  db:
+    image: postgres:12.3
+    container_name: nimble
```
rosle, line 6: Should we add `_db` suffix 📝

anduonghien: Fixed that in https://github.com/nimblehq/elixir-templates/pull/9/commits/3be8913e8e8017ff7d6ebf7e52b22266adef3dd0 @rosle

Issue #10, opened 2020-07-17
CLI Templatization


rosle:
## Feature request

- Shell script - https://github.com/infinitered/phoenix_base/blob/master/bin/rename_project
- https://github.com/cookiecutter/cookiecutter

rosle:
How to setup a README for the project. Right now the root README file is for the template. 📝

andyduong1920:
Close as we will implement Version 2.0

PR #14, opened 2020-08-07
CLI Templatization

Tolsee:
https://github.com/nimblehq/elixir-templates/issues/10

## What happened

Create a binary to rename project.
 
## Insight

 TBD

## Proof Of Work

TBD


bin/rename_project
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
anduonghien, line 1: as @olivierobert prefer to use Makefile instead of `bin/*` on my PR, should we made this with Makefile too 🤔 https://github.com/nimblehq/elixir-templates/pull/9#pullrequestreview-447245407

olivierobert: Makefile all the things ⚔️

Tolsee: @anduonghien @olivierobert 
I tried to look into the Makefile approach, but I think we will end up writing shell script inside the Makefile 😆 (Maybe because I am not very familiar writing Makefile). 

But, I think for this we can go with bash script because this is just a bootstrapping script i.e. IMO, it is not going to be part of a project itself it's just a part of the template. We can make this to remove itself after executing it. Also, I think we can move this script out of the bin(We can avoid introducing `/bin` to project created with this template). 

What do you think?

bin/rename_project
```diff
@@ -0,0 +1,28 @@
...
+mv lib/$CURRENT_OTP_NAME.ex lib/$2.ex
+mv lib/${CURRENT_OTP_NAME}_web.ex lib/$2_web.ex
+mv lib/$CURRENT_OTP_NAME lib/$2
+mv lib/${CURRENT_OTP_NAME}_web lib/$2_web
+[ -d test/${CURRENT_OTP_NAME} ] && mv test/${CURRENT_OTP_NAME} test/$2
+[ -d test/${CURRENT_OTP_NAME}_web ] && mv test/${CURRENT_OTP_NAME}_web test/$2_web
+
+echo "\n\nIf you cloned this template from Github, you may want to reinitialize git:"
+echo "    rm -rf .git/"
+echo ""
```
anduonghien, line 28: @Tolsee Please rename the docker container which is merged in https://github.com/nimblehq/elixir-templates/pull/9  too 👍

bin/rename_project
anduonghien: @Tolsee I think it's ok to write the shell script in the Makefile, like @rosle did in the Rails template

image

anduonghien: Also to make it easy for delete, we could name this as `init_app` and then we could delete that file after the job is done.

andyduong1920:
Close as we implement Version 2.0

PR #24, opened 2020-09-29
Adding integration test for Version 2.0

andyduong1920:

## What happened

Adding the integration test for Version 2.0
 
## Insight

We test the whole template by applying it to the new Phoenix project and running the test on the new project to make sure everything working on the new project.

This is the integration test, for the unit test for the template, will do it on another PR.

Test on several cases:
- Applying Web/API variant to a standard Phoenix project `mix phx.new NormalProject`
- Applying Web/API variant to a custom Phoenix project `mix phx.new CustomProject --app=CustomApp`
- Applying API variant to an Phoenix API project `mix phx.new ApiProject --no-html --no-webpack`

Pasted_Image_10_3_20__21_40

 
## Proof Of Work

https://github.com/nimblehq/elixir-templates/actions/runs/285914916


lib/nimble.phx.gen.template/addons/ex_coveralls.ex
```diff
@@ -13,9 +13,11 @@ defmodule Nimble.Phx.Gen.Template.Addons.ExCoveralls do
   end
 
   defp copy_files(%Project{} = project) do
+    # There is no test for `lib/*/router.ex` on the new Phoenix app, so that reduce the coverage to 77.8%
+    # We are setting the minimum_coverage as 77 to work with the new Phoenix app
```
andyduong1920, line 17: image

olivierobert: It looks kind of fragile, to be honest. What kind of tests would this require to keep it at 100%?

.github/workflows/new_app_test.yml
```diff
@@ -0,0 +1,108 @@
+name: "Generate new app"
```
olivierobert, line 1: Why do we need `test.yml` at all (I mean the other file)? I think this workflow could just replace it. So `new_app_test.yml` (which is not a great name 😆 ) would be `test.yml`, no?

lib/nimble.phx.gen.template/addons/ex_coveralls.ex
andyduong1920: We need to add the test for `api pipeline` and the LiveDashboard @olivierobert 

Screen_Shot_2020-10-01_at_11_08_08

andyduong1920: Fixed that in https://github.com/nimblehq/elixir-templates/pull/24/commits/519207309e21d483140a765a3412205d2ec48999 @olivierobert

Makefile
```diff
@@ -0,0 +1,16 @@
...
+	printf "Y\nY\n" | mix phx.new ${VARIANT}_project
+
+apply_template:
+	cd ${VARIANT}_project && \
+	echo '{:nimble_phx_gen_template, path: "../"},' > nimble_phx_gen_template.txt && \
+	sed -i -e '/{:phoenix, "~> /r nimble_phx_gen_template.txt' mix.exs && \
+	rm nimble_phx_gen_template.txt && \
+	mix deps.get && \
+	mix format && \
+	printf "Y\nY\nY\nY\n" | mix nimble.phx.gen.template --${VARIANT}
```
olivierobert, line 16: The 4 times `Y` is to accept the prompts, right? If there was more options, this build script would not work.

I found an alternative and we should be able to do:

```shell
yes Y | mix nimble.phx.gen.template --${VARIANT}
```

Check it out: https://en.wikipedia.org/wiki/Yes_(Unix)

andyduong1920: I tried the `yes command` already, that doesn't work @olivierobert as the Yes command will shutdown the shell process and the [Github job will stop](https://github.com/nimblehq/elixir-templates/runs/1181033071?check_suite_focus=true#step:10:57) at that moment.

https://stackoverflow.com/questions/20573282/hudson-yes-standard-output-broken-pipe

hanam1ni (approved):

Makefile
olivierobert: Noted. Too bad then.

.github/workflows/test.yml
```diff
@@ -1,8 +1,16 @@
 name: "Tests"
+
 on: pull_request
 
+env:
+  OTP_VERSION: 23.0.2
+  ELIXIR_VERSION: 1.10.4
+  PHOENIX_VERSION: 1.5.5
+  NODE_VERSION: 12
+  PROJECT_PATH: "sample_project"
```
olivierobert, line 10: This does not look like a path e.g. `/user/test/sample_project`. How about rename this to `PROJECT_DIRECTORY` or `PROJECT_ROOT`?

Makefile
olivierobert: Let's document what each `Y` corresponds to. Same for `create_project` as personally I don't recall what they are for. And your future self will not remember either in 6 months 😆

.github/workflows/test.yml
```diff
@@ -40,3 +48,123 @@ jobs:
         run: mix test
         env:
           MIX_ENV: test
+        
+  integration_test:
```
olivierobert, line 52: I am not sure why this is named `integration_test` as it executes the code formatter checks in each variant 🤔  Maybe naming it `test_variant` would be more descriptive? Also in that case, since both `unit_test` and `test_variant` are independent, it could be moved to another workflow file entirely -> `test_variant.yml`? It would make this file easier to manage in the long run IMO.

.github/workflows/test.yml
andyduong1920: How about `PROJECT_PATH: "/sample_project"` @olivierobert 🤔

andyduong1920: Fixed in https://github.com/nimblehq/elixir-templates/pull/24/commits/84a7e51b0a5abc1ddde3a9b0aee64f1c1d6700fb

.github/workflows/test.yml
andyduong1920: Fixed in https://github.com/nimblehq/elixir-templates/pull/24/commits/cf66116acbae3333aa7ce8de2fe068d41c51a19f

Makefile
andyduong1920: Added in https://github.com/nimblehq/elixir-templates/pull/24/commits/96e3f4d93baf339cc65c5a62fbd090729752a4fb @olivierobert

olivierobert (approved):
Looks good 👏

bterone (approved):

PR #26, opened 2020-10-02
Adding unit test for Version 2.0

andyduong1920:

## What happened

Adding the unit test for the template
 
## Insight

- Setup the AddonCase
- Adding unit test to all addons
 
## Proof Of Work

The [CI should be passed](https://github.com/nimblehq/elixir-templates/pull/26/checks?check_run_id=1207678648)

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,8 @@
 install_phoenix:
 	printf "Y\n" | mix archive.install hex phx_new ${PHOENIX_VERSION}
 
-# Y - in response to Override the existing ${PROJECT_DIRECTORY}?
-# Y - in response to Fetch and install dependencies?
 create_project:
-	printf "Y\nY\n" | mix phx.new ${PROJECT_DIRECTORY} ${OPTIONS}
+	mix phx.new ${PROJECT_PATH} ${OPTIONS}
 
 # Y - in response to Will you host this project on Github?
 # Y - in response to Do you want to generate the .github/ISSUE_TEMPLATE and .github/PULL_REQUEST_TEMPLATE?
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ install_phoenix:
 	printf "Y\n" | mix archive.install hex phx_new ${PHOENIX_VERSION}
 
 create_project:
-	mix phx.new ${PROJECT_PATH} ${OPTIONS}
+	mix phx.new ${PROJECT_DIRECTORY} ${OPTIONS}
 
 # Y - in response to Will you host this project on Github?
 # Y - in response to Do you want to generate the .github/ISSUE_TEMPLATE and .github/PULL_REQUEST_TEMPLATE?
--- a/lib/nimble.phx.gen.template/project.ex
+++ b/lib/nimble.phx.gen.template/project.ex
@@ -1,4 +1,6 @@
 defmodule Nimble.Phx.Gen.Template.Project do
+  alias Nimble.Phx.Gen.Template.Project
+
   defstruct otp_app: nil,
             base_module: nil,
             base_path: nil,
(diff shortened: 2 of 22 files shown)


test/nimble.phx.gen.template/addons/credo_test.exs
```diff
@@ -0,0 +1,55 @@
+defmodule Nimble.Phx.Gen.Template.Addons.CredoTest do
+  use Nimble.Phx.Gen.Template.AddonCase
+
+  setup %{project: project, test_project_path: test_project_path} do
```
hanam1ni, line 4: Just an idea 💡 How about we move this setup to `AddonCase` then make use of tags
```elixir
defmodule Nimble.Phx.Gen.Template.AddonCase do
  # ...

  setup context do
    # ...
    
    project = Nimble.Phx.Gen.Template.Project.info()

    if context[:apply_test_env] do
      File.cd!(test_project_path, fn ->
        Addons.TestEnv.apply(project)
      end)
    end

    {:ok, project: project, test_project_path: test_project_path}
  end
end
```

Then in test, we just add a tag and get rid of `setup` block 
```elixir
describe "#apply/2" do
    @tag :apply_test_env
    test "copies the .credo.exs" do
      # ...
    end

    @tag :apply_test_env
    test "injects credo to mix dependency" do
      # ...
    end
end
```

What do you think? 😄

test/nimble.phx.gen.template/addons/variants/web/sobelow_test.exs
```diff
@@ -0,0 +1,56 @@
+defmodule Nimble.Phx.Gen.Template.Addons.Web.SobelowTest do
+  use Nimble.Phx.Gen.Template.AddonCase
+
+  setup %{project: project, test_project_path: test_project_path} do
```
hanam1ni, line 4: If we refactor to tag as I mentioned above, we can apply Credo addon only test case that needed (apply only last one) 💭

hanam1ni (commented):
Only minor comments, the rest LGTM 👍

Makefile
```diff
@@ -4,7 +4,7 @@ install_phoenix:
 	printf "Y\n" | mix archive.install hex phx_new ${PHOENIX_VERSION}
 
 create_project:
-	printf "Y\nY\n" | mix phx.new ${PROJECT_PATH} ${OPTIONS}
+	mix phx.new ${PROJECT_PATH} ${OPTIONS}
```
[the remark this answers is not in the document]
andyduong1920: @hanam1ni On the unit test, we don't need to `Fetch and install the hex dependency`, but on the integration test, we need that.

lib/nimble.phx.gen.template/project.ex
```diff
@@ -7,4 +9,17 @@ defmodule Nimble.Phx.Gen.Template.Project do
             web_path: nil,
             web_test_path: nil,
             api_project?: nil
+
+  def info(opts \\ %{}) do
```
olivierobert, line 11: Controversial as it echoes OOP, but this method could also be named `new(opts \\ %{})` as it returns the Project module as a map.

test/nimble.phx.gen.template/addons/credo_test.exs
olivierobert: Good suggestion. Note that this `AddonCase` would be define the directory `/test`.

andyduong1920: @olivierobert I just following Phoenix project as they put the `xxxCase` in the `test/support` folder 
image

olivierobert: Yes. I noticed that afterwards and updated my comment above

andyduong1920: Actually, I prefer to keep the setup here, because that easy to read and know what we need to set up when we check the `credo_test` file, if we go with the tags, we need to find in the source code what and where we handle the `apply_test_env` there.

And think about if we have another case like this, we also put to the AddonCase, that will make the AddonCase bigger.

Another thing, in this test file, we have only 2 test cases, so we can easily put the `@tag :apply_test_env` 2 times, but if we have 10 test cases, we need to put the `@tag :apply_test_env` 10 times ☁️

andyduong1920: How about this implementation @hanam1ni @olivierobert https://github.com/nimblehq/elixir-templates/pull/26/commits/8a8ec445839df370405291d770c1b078fe74d777

hanam1ni: looks good to me 🙌 . How about rename from `presetup` to make it less generic and more descriptive,  maybe `base_addons` or `apply_addons` (feel free to use others name 😄 ). 

Approving the PR 👍

hanam1ni (approved):

test/nimble.phx.gen.template/addons/credo_test.exs
andyduong1920: or `pre_apply` or `required_addons` 🤔  @olivierobert do you have any suggestion 😄

andyduong1920: Changed it to `required_addons` @hanam1ni 😃

olivierobert (approved):

bterone (approved):

PR #27, opened 2020-10-05
Fetch the latest Hex package version

andyduong1920:

## What happened

Fetch the latest hex package version
 
## Insight

Instead of hard code the version we could fetch the latest version when the template is trigger.

This PR solving 2 problems:

- Reduce the maintenance effort as we no need to increase the version manually, open a PR, merge, release,...
- Always fetch the latest version when applying the template
 
I don't use the [Application environment ](https://hexdocs.pm/elixir/Application.html)for this piece of code because that [isn't the best practice for a library ](https://hexdocs.pm/elixir/library-guidelines.html#avoid-application-configuration)

```
      defp hex_package(:test), do: Nimble.Phx.Gen.Template.Hex.PackageMock
      defp hex_package(_env), do: Nimble.Phx.Gen.Template.Hex.Package
```

## Proof Of Work

- The unit test suite should be passed (thanks to Mox)
- The integration test also passed


olivierobert (commented):
Looks good. I would still prefer to remove the `setup` block as discussed in #26. It would remove the coupling in all tests.

andyduong1920:
> Looks good. I would still prefer to remove the `setup` block as discussed in #26. It would remove the coupling in all tests.

It doesn't remove actually, it makes the setup as hidden by moving to the AddonCase, so the test will look magical IMO.

hanam1ni (approved):

andyduong1920:
> Looks good. I would still prefer to remove the `setup` block as discussed in #26. It would remove the coupling in all tests.

I removed the setup block and adding the `@describetag mock_latest_package_versions: [{:xx, "xx"}]` tag already ;)

olivierobert (approved):

bterone (approved):

Issue #28, opened 2020-10-06
Adding Credo to the template itself


andyduong1920:
## Feature request

Adding Credo to the Template codebase itself.

PR #29, opened 2020-10-06
Adding credo

andyduong1920:
https://github.com/nimblehq/elixir-templates/issues/28

## What happened

Solve #28 
 
## Insight

Adding Credo to the template itself
 
## Proof Of Work

image


olivierobert (approved):

hanam1ni (approved):

.credo.exs
```diff
@@ -0,0 +1,175 @@
...
+        {Credo.Check.Consistency.SpaceInParentheses, []},
+        {Credo.Check.Consistency.TabsOrSpaces, []},
+
+        #
+        ## Design Checks
+        #
+        # You can customize the priority of any check
+        # Priority values are: `low, normal, high, higher`
+        #
+        {Credo.Check.Design.AliasUsage, false},
```
bterone, line 75: I think here can we use the 
```
{Credo.Check.Design.AliasUsage,
         [priority: :low, if_nested_deeper_than: 2, if_called_more_often_than: 0]},
``` 
I got it from the [credo github](https://github.com/rrrene/credo/blob/master/.credo.exs).
On GCP we use `if_called_more_often_than: 3` instead.
We can also customize it depending on the project, so I don't mind either and we can keep it like this 😂

bterone (approved):

.credo.exs
andyduong1920: Thanks @bterone let me check that again

bterone (approved):

.credo.exs
andyduong1920: Fixed that in https://github.com/nimblehq/elixir-templates/pull/29/commits/420691a58cd6cac2328e90b39116ce2cb8f237a6 :)

PR #36, opened 2020-10-13
Support Elixir 1.11

andyduong1920:
## What happened

Support Elixir 1.11
 
## Insight

- Require the new Phoenix project update to the new Elixir and OTP version (to support different Elixir and OTP version, we could add that to the roadmap)

- Update the README
 
## Proof Of Work

The test is passed on CI


lib/nimble.phx.gen.template/addons/elixir_version.ex
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  defp edit_files(project) do
+    Generator.replace_content(
+      "mix.exs",
+      """
+            elixir: "~> 1.7",
+      """,
+      """
+            elixir: "~> 1.11",
```
olivierobert, line 24: What do you think of having a constant defined in `Nimble.Phx.Gen.Template.Project` 💭 The advantage is that it would not be hidden deep in an addon.

andyduong1920: Moved to Project in https://github.com/nimblehq/elixir-templates/pull/36/commits/ba94b5ddb712ae8475daec36ff238b86460e1b87 @olivierobert

olivierobert (approved):

hanam1ni (approved):

Issue #12, opened 2020-07-17
Dockerize the app


rosle:
## Feature request

- Add dockerfile 
- Add docker-compose for dev/production
- Check how the container deployment work

- Create `release.exs` and remove `prod.secret.exs`

anduonghien:
https://hexdocs.pm/phoenix/releases.html#containers

PR #37, opened 2020-10-13
Adding Mix Release with Docker

andyduong1920:
Solve https://github.com/nimblehq/elixir-templates/issues/12

## What happened

Dockerize the App with Mix Release
 
## Insight

- Update `config/prod.exs`
- Create the `config/runtime.exs`
- Adding `Dockerfile, docker-compose.yml`
- Adding the `bin/start.sh` for the Docker entry point, I decided to put that in the `bin/` folder as because after the app is built, it has the `bin/` folder there.

image

Refer: [Phoenix release](https://hexdocs.pm/phoenix/releases.html)

- Adding Docker build script on Github workflow to test the Build image and app scripts

## Proof Of Work

https://drive.google.com/file/d/1rsOeWJBmAvvIC6uej9JTkpQFoSM51KJf/view

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -170,7 +170,7 @@
         {Credo.Check.Refactor.DoubleBooleanNegation, []},
         {Credo.Check.Refactor.ModuleDependencies, false},
         {Credo.Check.Refactor.NegatedIsNil, []},
-        {Credo.Check.Refactor.PipeChainStart, false},
+        {Credo.Check.Refactor.PipeChainStart, []},
         {Credo.Check.Refactor.VariableRebinding, []},
         {Credo.Check.Warning.LeakyEnvironment, []},
         {Credo.Check.Warning.MapGetUnsafePass, []},
--- a/.credo.exs
+++ b/.credo.exs
@@ -170,7 +170,7 @@
         {Credo.Check.Refactor.DoubleBooleanNegation, []},
         {Credo.Check.Refactor.ModuleDependencies, false},
         {Credo.Check.Refactor.NegatedIsNil, []},
-        {Credo.Check.Refactor.PipeChainStart, []},
+        {Credo.Check.Refactor.PipeChainStart, false},
         {Credo.Check.Refactor.VariableRebinding, []},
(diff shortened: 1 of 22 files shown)


.credo.exs
```diff
@@ -170,7 +170,7 @@
         {Credo.Check.Refactor.DoubleBooleanNegation, []},
         {Credo.Check.Refactor.ModuleDependencies, false},
         {Credo.Check.Refactor.NegatedIsNil, []},
-        {Credo.Check.Refactor.PipeChainStart, []},
+        {Credo.Check.Refactor.PipeChainStart, false},
```
olivierobert, line 173: Is this really needed as [per the example](https://github.com/rrrene/credo/blob/master/lib/credo/check/refactor/pipe_chain_start.ex), this is actually nicer to read

priv/templates/nimble.phx.gen.template/Dockerfile.eex
```diff
@@ -0,0 +1,49 @@
...
+
+# prepare release image
+FROM alpine:3.12.0 AS app
+RUN apk add --no-cache openssl ncurses-libs
+
+WORKDIR /app
+
+RUN chown nobody:nobody /app
+
+USER nobody:nobody
```
olivierobert, line 42: Not sure if `nobody` is a standard non-root user name and group but maybe having something more descriptive would be better

olivierobert (commented):
Looks good

.credo.exs
andyduong1920: Reverted in https://github.com/nimblehq/elixir-templates/pull/37/commits/8ef2ac4d2dfcb4ac4f9bd1a657aea819c847fdb2

priv/templates/nimble.phx.gen.template/Dockerfile.eex
andyduong1920: changed it to `app_group` and `app_user` @olivierobert

olivierobert (approved):

hanam1ni (approved):

PR #55, opened 2020-11-19
Replace Mox to Mimic

andyduong1920:
## What happened

Replacing Mox to Mimic
 
## Insight

As Mimic is simpler, requiring less configuration like Mox
 
## Proof Of Work

The test still pass


olivierobert (approved):
I am fine with using Mimic and there is nothing wrong with this PR 😁 I still think that using the architecture of behaviours for Mock has its place. Even if we plan to use Mimic to simplifying testing, I would recommend using the env-based definition of modules (as explained in [the Plataformatec original article](http://blog.plataformatec.com.br/2015/10/mocks-and-explicit-contracts/)) for critical modules or third-party client wrapper modules.

bterone (approved):

Issue #16, opened 2020-08-10
Adding `Project README`


andyduong1920:
## Why

Once we generate the project, we need some `READ ME` like how to set up, install the dependency, running test,...

## Feature request

Create a `PROJECT_README.md` file for example, that contains some content like

- Setup docker
- Setup chromedriver,...
- How to run test,...

And in the #14 we could remove the `README.md` (template readme) and rename the PROJECT_README.md to just README.md so the new project will have some initial useful README content

PR #56, opened 2020-11-20
Add Project README file

andyduong1920:
## What happened

https://github.com/nimblehq/elixir-templates/issues/16
 
## Insight

Adding the Readme Addon
 
## Proof Of Work

image

### DIFF
--- a/priv/templates/nimble.phx.gen.template/README.md.eex
+++ b/priv/templates/nimble.phx.gen.template/README.md.eex
@@ -14,7 +14,7 @@
 
   - [asdf](https://github.com/asdf-vm/asdf) Erlang & Elixir
 
-### Docker
+### Development
 
 * Install [Docker for Mac](https://docs.docker.com/docker-for-mac/install/)
 
@@ -23,8 +23,6 @@
   ```sh
   make docker_setup
   ```
-
-### Development
 
 * Install Elixir dependencies:
 
@@ -50,15 +48,11 @@
   iex -S mix phx.server
   ```
 
-### Test
-
 * Run all tests:
 
   ```sh
   mix test 
   ```
-
-### Code Format
 
 * Run all lint:
 
@@ -66,8 +60,10 @@
   mix codebase 
   ```
 
-### Build Docker Production Image
+### Production
 
-```sh
-docker-compose build
-```
+* Buidl Docker image
+
+  ```sh
+  docker-compose build
+  ```
--- a/priv/templates/nimble.phx.gen.template/README.md.eex
+++ b/priv/templates/nimble.phx.gen.template/README.md.eex
@@ -31,26 +31,23 @@
   ```sh
   mix deps.get
   ```
-
+<%= if web_project? do %>
 * Install Node dependencies:
 
   ```sh
   npm install --prefix assets
   ```
-
+<% end %>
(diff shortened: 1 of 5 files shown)


priv/templates/nimble.phx.gen.template/README.md.eex
```diff
@@ -0,0 +1,69 @@
...
+
+## Project Setup
+
+### Erlang & Elixir
+
+* Erlang <%= erlang_version %>+ and Elixir <%= elixir_version %>+
+
+* Recommended version manager.
+
+  - [asdf](https://github.com/asdf-vm/asdf) Erlang & Elixir
```
olivierobert, line 15: Controversial but I would not recommend it just yet 😛 . There is discussion about it on [Elixir Forums](https://elixirforum.com/t/elixir-version-managers/317/5):

> asdf has the big downsite, that it hides the installed system version completely and you can’t use it except by fully qualifying it (might create other problems because system and asdf executables are called in a mix) or kicking asdfs binary wrapper folder out of the $PATH

I experienced that first hand with Jetbrains-based IDE and I think it can cause issues in other cases.

Maybe let's leave this out for now and at the same time maybe we should check out [kiex](https://github.com/taylor/kiex)

test/nimble.phx.gen.template/addons/readme_test.exs
```diff
@@ -0,0 +1,55 @@
...
+    test "copies the README.md", %{
+      project: project,
+      test_project_path: test_project_path
+    } do
+      in_test_project(test_project_path, fn ->
+        Addons.Readme.apply(project)
+
+        assert_file("README.md", fn file ->
+          assert file =~ """
+                 Erlang 23.1.1+ and Elixir 1.11+
```
olivierobert, line 14: Not sure to add these checks as it will break on each update

andyduong1920: yes, it will fail when we `upgrade the Erlang version or Elixir version` for the template but that is what we want to have thus we would like to make sure the README is up to date along with the new version too, so we need to adjust the version in this test too :)

priv/templates/nimble.phx.gen.template/README.md.eex
andyduong1920: asdf work well with me :D never use [kiex](https://github.com/taylor/kiex) before, will have a look on that ;)

andyduong1920: I just have had a look on kiex and the first thing I saw is that doesn't support erlang, so we end up using 2 different version manage tool (1 for erlang and kiex for elixir) 🙃 .

Pasted_Image_27_11_2020__09_47

bterone (approved):

priv/templates/nimble.phx.gen.template/README.md.eex
andyduong1920: I also found an article comparing asdf and kiex and other tools https://nts.strzibny.name/elixir-version-manager/

olivierobert: Alright, I will give `asdf` a chance again 😁

hanam1ni (approved):

Issue #42, opened 2020-11-12
Core JS addon


andyduong1920:
https://www.npmjs.com/package/core-js

PR #58, opened 2020-11-20
Adding CoreJS Addon

andyduong1920:
## What happened

Solved #42 
 
## Insight

https://github.com/zloirock/core-js#installation
 
## Proof Of Work

The test is passed

Tested on my local

image

Pasted_Image_20_11_2020__16_58

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,13 @@ create_project:
 # Y - in response to Do you want to generate the Github Action workflow?
 # Y - in response to Would you like to add the Oban addon?
 # Y - in response to Fetch and install dependencies?
+common_addon_prompts = "Y\nY\nY\nY\nY\n"
+
+# Y - in response to Would you like to add the CoreJS addon?
+web_addon_prompts = "Y\n"
+
+api_addon_prompts = ""
+
 apply_template:
 	cd ${PROJECT_DIRECTORY} && \
 	echo '{:nimble_phx_gen_template, path: "../", only: :dev, runtime: false},' > nimble_phx_gen_template.txt && \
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -11,15 +11,28 @@ create_project:
...
 # Y - in response to Would you like to add the Oban addon?
+common_addon_prompts = Y\nY\nY\nY\n
+
+# Y - in response to Would you like to add the CoreJS addon?
+web_addon_prompts = Y\n
+
+api_addon_prompts = 
+
 # Y - in response to Fetch and install dependencies?
+suffix_addon_prompts = Y\n
```
olivierobert, line 22: `suffix` is kinf of awkward here 🤔 How about `post_setup_addon_prompts. This would be similar to [NPM](https://docs.npmjs.com/cli/v6/using-npm/scripts):

Screen Shot 2563-11-26 at 16 59 36

Makefile
```diff
@@ -11,15 +11,28 @@ create_project:
 # Y - in response to Do you want to generate the .github/ISSUE_TEMPLATE and .github/PULL_REQUEST_TEMPLATE?
 # Y - in response to Do you want to generate the Github Action workflow?
 # Y - in response to Would you like to add the Oban addon?
+common_addon_prompts = Y\nY\nY\nY\n
+
+# Y - in response to Would you like to add the CoreJS addon?
```
olivierobert, line 16: Is there really a case when CoreJS would not be needed? Shouldn't bt installed by default for a Web project since all projects would use JS?

andyduong1920: Actually, I have had the same thoughts too, haha, will set it to default addon for Web variant 👍

andyduong1920: Fixed in https://github.com/nimblehq/elixir-templates/pull/58/commits/4785e4f3f64137f8ba4cfc0d5f8cb8bbc25ece81

Makefile
andyduong1920: Fixed in https://github.com/nimblehq/elixir-templates/pull/58/commits/fa6253fd3a29280ed91c3b6d0d457e075ef04df4

olivierobert (approved):

hanam1ni (approved):

PR #60, opened 2020-11-22
Adding LiveView variant

andyduong1920:
## What happened

Solved https://github.com/nimblehq/elixir-templates/issues/59

The template could support the Phoenix LiveView project.
 
## Insight

1/ Adding the `--live` option
2/ Refactor the variant a bit, to make it clear between `web/api/live` project
3/ Skip the Code Cover All on the `page_live.ex` as that is the default screen from Phoenix and their don't add a full test for that module.
4/ Test for LiveView Project on Github Action

## Proof Of Work

1/ Tested on my local
2/ Test is passed on Github Action

### DIFF
--- a/.github/workflows/test_live_variant_on_custom_liveview_project.yml
+++ b/.github/workflows/test_live_variant_on_custom_liveview_project.yml
@@ -3,10 +3,10 @@ name: "Test Live variant on custom LiveView project"
 on: push
 
 env:
-  OTP_VERSION: 23.1.1
-  ELIXIR_VERSION: 1.11.1
-  PHOENIX_VERSION: 1.5.5
-  NODE_VERSION: 12
+  OTP_VERSION: 23.1.4
+  ELIXIR_VERSION: 1.11.2
+  PHOENIX_VERSION: 1.5.7
+  NODE_VERSION: 14
 
 jobs:   
   unit_test:
--- a/.github/workflows/test_live_variant_on_custom_liveview_project.yml
+++ b/.github/workflows/test_live_variant_on_custom_liveview_project.yml
@@ -6,6 +6,7 @@ env:
   OTP_VERSION: 23.1.1
   ELIXIR_VERSION: 1.11.1
   PHOENIX_VERSION: 1.5.5
+  NODE_VERSION: 12
 
 jobs:   
   unit_test:
@@ -32,6 +33,10 @@ jobs:
         with:
           otp-version: ${{ env.OTP_VERSION }}
           elixir-version: ${{ env.ELIXIR_VERSION }}
+        
+      - uses: actions/setup-node@v2-beta
+        with:
+          node-version: ${{ env.NODE_VERSION }}
           
       - name: Cache Elixir build
         uses: actions/cache@v2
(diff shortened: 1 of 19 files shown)


.github/workflows/test_live_variant_on_custom_liveview_project.yml
```diff
@@ -0,0 +1,69 @@
+name: "Test Live variant on custom LiveView project"
```
olivierobert, line 1: Having `live` both seems redundant. The following would be sufficient and would be consistent with the other workflows:

```suggestion
name: "Test variant on custom LiveView project"
```

.github/workflows/test_live_variant_on_custom_liveview_project.yml
```diff
@@ -0,0 +1,69 @@
...
+
+on: push
+
+env:
+  OTP_VERSION: 23.1.1
+  ELIXIR_VERSION: 1.11.1
+  PHOENIX_VERSION: 1.5.5
```
olivierobert, line 8: `test_variant_on_custom_project.yml` and `test_variant_on_standard_project.yml` use different versions:

```
OTP_VERSION: 23.1.4
ELIXIR_VERSION: 1.11.2
PHOENIX_VERSION: 1.5.7
````

Any reasons? If not then it should be synced with these versions.

File: lib/nimble.phx.gen.template/project.ex [hunk 3]
```diff
@@ -29,14 +31,26 @@ defmodule Nimble.Phx.Gen.Template.Project do
...
+      web_path: "lib/" <> Atom.to_string(otp_app()) <> "_web",
+      web_test_path: "test/" <> Atom.to_string(otp_app()) <> "_web"
     }
   end
+
+  defp api_project?(opts), do: opts[:api] === true
+
+  defp web_project?(opts), do: opts[:web] === true || opts[:live] === true
+
+  defp live_project?(opts), do: opts[:live] === true
```
olivierobert, line 51: First adding another param/opts such as `web`, `api` and `live` is rather confusing as it's not an attribute of `Nimble.Phx.Gen.Template.Project`. So there is some indirectionr there.

Second, I think we should try `live` and `api` independently. It's possible for an API to have some live views. It's actually the case in Omise Labs as the integration with third-party payment providers requires to implements some views for redirects. To make things flexible, we should allow for both `web` and `api` projects to have LiveView enabled. So `LiveView` could be treated as an addon instead of a variant.

I understand it would be a major change but I think it's the right approach.

.github/workflows/test_live_variant_on_custom_liveview_project.yml
andyduong1920: yeah, good catch, the new version is merged before this branch and I forget to adjust this branch.

lib/nimble.phx.gen.template/project.ex
andyduong1920: An API project having the live view, that is not standard @olivierobert as the API project here means that project is generated by `mix phx.new name --no-html --no-webpack`, with no HTML and Webpack....

In Omise, they created a `Web project` and having only `API` endpoints, that is different with a `no HTML, Webpack` project 😀

We can't generate a `live` project along with `--no-html --no-webpack`

image

hanam1ni (approved):

.github/workflows/test_live_variant_on_custom_liveview_project.yml
andyduong1920: It follows the API variant @olivierobert :) 

image

image

.github/workflows/test_live_variant_on_custom_liveview_project.yml
andyduong1920: Upgraded in https://github.com/nimblehq/elixir-templates/pull/60/commits/6f3ec24e5493d82b65e13424b8f3a120a11e861d

lib/nimble.phx.gen.template/project.ex
olivierobert: Okay, I understand and you are right that Omise Labs created a Web project. It's just that I am not fully satisfied with having the current options as it's not that simple to understand all of them. I would prefer to find ways this can be simplified.

olivierobert (approved):
The implementation of Live Views looks good to me. I am having concerns on the complexity added to the configuration.

Issue #68, opened 2020-12-03
Unit test to verify the `version in mix.exs file` on each `release/* branches` is greater than the version on hex.pm


andyduong1920:
To make sure we always increasing the version `either major or minor` on each release.

PR #71, opened 2020-12-18
Validate the New Version to be Published on each Release branch

andyduong1920:
## What happened

Setup a test to check the new version to be published on each`release branches` should be greater than the `hex` version.

Solved https://github.com/nimblehq/elixir-templates/issues/68
 
## Insight

As the version number in the `mix.ex` file needs to be increased before releasing to Hex.pm **manually** so sometimes the developer could forget todo that.

In case the new version is the same with the hex version lower than it, hex.pm doesn't allow to publish the new version
 
## Proof Of Work

Try to trigger that workflow

image

### DIFF
--- a/test/nimble.phx.gen.template/release_version_test.exs
+++ b/test/nimble.phx.gen.template/release_version_test.exs
@@ -1,0 +1,13 @@
+defmodule Nimble.Phx.Gen.Template.ReleaseVersionTest do
+  use ExUnit.Case, async: true
+
+  alias Nimble.Phx.Gen.Template.Hex.Package
+
+  @tag :release_version
+  test "the new version should be greater than the hex version" do
+    new_version = Mix.Project.config()[:version]
+    hex_version = Package.get_latest_version("nimble_phx_gen_template")
+
+    assert new_version > hex_version
+  end
+end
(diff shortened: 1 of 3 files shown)


test/nimble.phx.gen.template/release_version_test.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Nimble.Phx.Gen.Template.ReleaseVersionTest do
+  use ExUnit.Case, async: true
+
+  alias Nimble.Phx.Gen.Template.Hex.Package
+
+  @tag :release_version
+  test "the new version should be greater than the hex version" do
```
olivierobert, line 7: Prefer the imperative tone in tests 🙈 

```suggestion
  test "the new version is greater than the hex version" do
```

olivierobert (changes requested):

test/nimble.phx.gen.template/release_version_test.exs
andyduong1920: Fixed in https://github.com/nimblehq/elixir-templates/pull/71/commits/c0e6ac5f2c7d453c82d1b1073354c10dba1e378f @olivierobert ✅

hanam1ni (approved):

rosle (approved):

olivierobert (approved):

PR #79, opened 2021-01-08
Refactor the default version component

andyduong1920:
## What happened

Simplify the `default versions` so, in the future, we just need to update either `@elixir_version` or `@erlang_version` or `@alpine_version`
 
## Insight

1/ Simplify the `default versions` so, in the future, we just need to update either `@elixir_version` or `@erlang_version` or `@alpine_version`
2/ Refactor the Github Action Elixir, Erlang and Node version
3/ Upgrade Elixir to 1.11.4 and Erlang to 23.3
 
## Proof Of Work

The test is GREEN on CI

### DIFF
--- a/lib/nimble.phx.gen.template/project.ex
+++ b/lib/nimble.phx.gen.template/project.ex
@@ -1,17 +1,8 @@
 defmodule Nimble.Phx.Gen.Template.Project do
-  # Erlang versions: asdf list all erlang
-  # Elixir versions: asdf list all elixir
-  @default_versions %{
-    erlang_asdf_version: "23.2.1",
-    elixir_asdf_version: "1.11.3-otp-23",
-    elixir_mix_version: "1.11.3"
-  }
+  @elixir_version "1.11.3"
+  @erlang_version "23.2.1"
 
-  # Elixir image tags: https://hub.docker.com/r/hexpm/elixir/tags
-  @docker_base_images %{
-    build: "hexpm/elixir:1.11.3-erlang-23.2.1-alpine-3.12.1",
-    app: "alpine:3.12.1"
-  }
+  @alpine_version "3.12.1"
 
   defstruct otp_app: nil,
             base_module: nil,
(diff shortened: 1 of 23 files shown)


lib/nimble.phx.gen.template/project.ex
```diff
@@ -23,11 +15,14 @@ defmodule Nimble.Phx.Gen.Template.Project do
...
-            docker_build_base_image: @docker_base_images[:build],
-            docker_app_base_image: @docker_base_images[:app]
+            elixir_version: @elixir_version,
+            erlang_version: @erlang_version,
+            node_version: @node_version,
+            elixir_asdf_version:
+              "#{@elixir_version}-otp-#{@erlang_version |> String.split(".") |> List.first()}",
+            docker_build_base_image:
+              "hexpm/elixir:#{@elixir_version}-erlang-#{@erlang_version}-alpine-#{@alpine_version}",
+            docker_app_base_image: "alpine:#{@alpine_version}"
```
olivierobert, line 25: Since there are many many atributes now 😆 do you think it would be better to order them alphabetically or have groupings:

```
defstruct otp_app: nil,
          # Dependency Versions 
          elixir_version: @elixir_version,
          erlang_version: @erlang_version,
          node_version: @node_version,
          elixir_asdf_version: "#{@elixir_version}-otp-#{@erlang_version |> String.split(".") |> List.first()}",
          # Variants
          api_project?: false,
          web_project?: false,
          live_project?: false,
```

Also having documentation for attributes like `otp_app` and `base_module` to better understand what this does would be good. This can be added in the moduledoc in another PR.

priv/templates/nimble.phx.gen.template/.github/workflows/test.yml.eex
```diff
@@ -30,7 +32,11 @@ jobs:
         with:
           otp-version: ${{ env.OTP_VERSION }}
           elixir-version: ${{ env.ELIXIR_VERSION }}
-
+      <%= if web_project? do %>
+      - uses: actions/setup-node@v2-beta
```
olivierobert, line 36: The stable version is [now released](https://github.com/actions/setup-node/releases/tag/v2.1.5):

```suggestion
      - uses: actions/setup-node@v2.1.5
```

olivierobert (commented):
It's definitely an improvement 👍

priv/templates/nimble.phx.gen.template/.github/workflows/test.yml.eex
andyduong1920: Thanks, I have fixed in https://github.com/nimblehq/elixir-templates/pull/79/commits/47035e61fc0877040fe376ca2312223eef025146

lib/nimble.phx.gen.template/project.ex
andyduong1920: yeah 👍 I have fixed it in https://github.com/nimblehq/elixir-templates/pull/79/commits/e05f49bae25ec883d6d325b2b8576b41e1f88678 :)

byhbt (approved):
LGTM 🤖

hanam1ni (approved):

PR #64, opened 2020-11-27
Adding Mix variant

andyduong1920:
## What happened

Adding Mix variant #33 
 
## Insight

1/ Adding the `@moduledoc` into the mix task

2/ Supporting `--mix` option

3/ Supported addons

- ElixirVersion
- Readme
- TestEnv
- Credo
- Dialyxir
- ExCoveralls
- Github template and Github action workflow (optional)
- Mimic (optional)

4/ The mix task still as `nimble.phx.gen.template` so far, will rename the mix task on another PR.
 
## Proof Of Work

1/ Adding the `mix help nimble.phx.gen.template`

image

2/ The test is passed

3/ [Demo](https://drive.google.com/file/d/1ZwnT943WaJykga2svPAASZZP8yCaOp26/view?usp=sharing)

Issue #33, opened 2020-10-12
Mix project variant?


andyduong1920:
Currently, the template is supporting the Phoenix project (API and web variants), how about supporting the Mix project (for creating another Elixir lib) 🤔

```
mix nimble.phx.gen.template --mix
```

We could consider changing the task name to `nimble.gen.template` so that doesn't contain the `phx` to make it clear

```
mix nimble.gen.template --mix
```

olivierobert:
That would be a great addition as even for umbrella apps, we might not always need a Phoenix application. So I am in favour or renaming the task name to `mix nimble.gen.template`.

PR #81, opened 2021-01-11
Rename the project into `nimble_template` and the mix task into `nimble.template.gen`

andyduong1920:
## What happened

As the template has supported `mix variant` (#64), it's better to rename the `mix task` by removing the `phx` to make it clearer (following the discussion in #33)
 
## Insight

1/ Changing the mix task from `mix nimble.phx.gen.template` into `mix nimble_template.gen`
2/ Changing the folder from `nimble.phx.gen.template` into `nimble_template` (having the dots on folder name isn't Elixir convention)
3/ Changing the Elixir's modules prefix from `Nimble.Phx.Gen.Template` to just `NimbleTemplate` to following [Elixir community's package name convention](https://hex.pm/docs/faq) 

image

4/ Adding `@moduledoc false` to clarify that module intent to not have doc :) 

5/ Rename the package from `nimble_phx_gen_template` to just `nimble_template` (a new package)

6/ Disable the `Test Release version` Github Action workflow as we gonna release a new `hex package` `nimble_template` so at the first time there is no `previous` version to check. We will enable that workflow on the next release.

7/ The package name convention is following `https://github.com/dashbitco/nimble_parsec` 

8/ Bump to 3.0.0

## TODO

- [ ] Set the `https://hex.pm/packages/nimble_phx_gen_template` as retired (as it's the old package and Hex doesn't allow us to delete a package - https://hex.pm/docs/faq)

## Proof Of Work

1/ The test is passed

2/ The mix task is still working as normal
image


olivierobert (approved):

hanam1ni (approved):
👍  Changes are LGTM. One small thing, could you update a PR description a bit, adding a reason why we mark `@moduledoc` as false?

andyduong1920:
> Changes are LGTM 👍
> 
> One small thing, could you update a PR description a bit, adding a reason why we mark `@moduledoc` as false?

Thanks for your feedback @hanam1ni I have updated the PR's description :)

PR #84, opened 2021-03-17
Change the first app name in mix phx.new to lower key

byhbt:
## What happened

As the warning from the `mix` command when we try to create a new app with the **first upper case** char.

Screen Shot 2021-03-17 at 10 09 12 AM

## Insight

n/a
 
## Proof Of Work

n/a


andyduong1920 (approved):

bterone:
Should it actually be in `snake_case`? @byhbt In the [examples in the docs](https://hexdocs.pm/phoenix/Mix.Tasks.Phx.New.html#module-examples), it shows it in snake case. 

It also mentions in the error it should only contain lower case characters. 🤔

byhbt:
> Should it actually be in `snake_case`? @byhbt
> In the [example in the docs](https://hexdocs.pm/phoenix/Mix.Tasks.Phx.New.html#module-examples), it shows app name in snake case.
> 
> It also mentions in the error it should only contain lower case characters. 🤔

hi @bterone !

I changed the name from `helloElixirApp` to `hello_elixir` seems more appropriate. 

Do you have any other suggestion? :) 

Thanks for the good catch!

olivierobert (approved):

PR #90, opened 2021-04-05
Adding ExVCR addon

hanam1ni:
## What happened

As mentioned in [Compass](https://nimblehq.co/compass/development/code-conventions/ex-unit/#stubbing-network-requests), we going to use ExVCR for stubbing network request in tests. In this PR, adding ExVCR addon:

- Adding ExVCR dependency. 
- Update ExVCR configuration. 
- Creating a directory for cassettes.
 
## Insight

Adding 2 helper functions in `Nimble.Phx.Gen.Template.Generator` module

- `append_content/2` for appending content to the given file.
- `make_directory/2` for creating directory and providing an option to create a `.keep` file.
 
## Proof Of Work

After applied templated to the project:

- ExVCR is added in `deps` of `mix.exs`
image

- Configuration is updated in `config/test.exs`
image

- `DataCase` and `ConnCase` are updated
image

- Create a dummy module and add a test. After running the test, a cassette is created.
image

### DIFF
--- a/lib/nimble.phx.gen.template/addons/ex_vcr.ex
+++ b/lib/nimble.phx.gen.template/addons/ex_vcr.ex
@@ -1,0 +1,73 @@
+defmodule Nimble.Phx.Gen.Template.Addons.ExVCR do
+  use Nimble.Phx.Gen.Template.Addon
+
+  @cassette_directory "test/support/fixtures/vcr_cassettes"
+
+  @impl true
+  def do_apply(%Project{} = project, _opts) do
+    project
+    |> edit_files()
+    |> create_cassette_directory()
+  end
+
+  defp edit_files(%Project{} = project) do
+    project
+    |> inject_mix_dependency()
+    |> edit_test_config()
+    |> edit_case()
+
+    project
+  end
+
+  defp inject_mix_dependency(%Project{} = project) do
+    Generator.inject_mix_dependency({:exvcr, latest_package_version(:exvcr), only: :test})
+
+    project
+  end
+
+  defp edit_test_config(project) do
+    Generator.append_content(
+      "config/test.exs",
+      """
+
+      # Configurations for ExVCR
+      config :exvcr,
+        vcr_cassette_library_dir: "#{@cassette_directory}"
+      """
+    )
+
+    project
+  end
+
+  defp edit_case(project) do
+    Generator.inject_content(
+      "test/support/conn_case.ex",
+      """
+          quote do
+      """,
+      """
+            use ExVCR.Mock, adapter: ExVCR.Adapter.Hackney
+
+      """
+    )
+
+    Generator.inject_content(
+      "test/support/data_case.ex",
+      """
+          quote do
+      """,
+      """
+            use ExVCR.Mock, adapter: ExVCR.Adapter.Hackney
+
+      """
+    )
+
+    project
+  end
+
+  defp create_cassette_directory(project) do
+    Generator.make_directory(@cassette_directory)
+
+    project
+  end
+end
--- a/test/nimble.phx.gen.template/addons/ex_vcr_test.exs
+++ b/test/nimble.phx.gen.template/addons/ex_vcr_test.exs
@@ -1,0 +1,62 @@
+defmodule Nimble.Phx.Gen.Template.Addons.ExVCRTest do
+  use Nimble.Phx.Gen.Template.AddonCase
+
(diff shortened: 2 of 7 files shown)


lib/nimble.phx.gen.template/addons/ex_vcr.ex
```diff
@@ -0,0 +1,73 @@
+defmodule Nimble.Phx.Gen.Template.Addons.ExVCR do
+  use Nimble.Phx.Gen.Template.Addon
+
+  @cassette_directory "test/support/fixtures/vcr_cassettes"
+
+  @impl true
+  def do_apply(%Project{} = project, _opts) do
```
byhbt, line 7: I have a question here about this variable `_opts`, how to deal with the unused variable? if we don't use it should we rename it to `_` ? 🤔

lib/nimble.phx.gen.template/generator.ex
```diff
@@ -86,4 +111,16 @@ defmodule Nimble.Phx.Gen.Template.Generator do
...
     end
   end
+
+  defp create_keep_file(path, true) do
+    case File.touch("#{path}/.keep") do
+      :ok ->
+        :ok
+
+      {:error, reason} ->
+        Mix.raise(~s[Failed to create keep file at #{path} reason: #{Atom.to_string(reason)}])
```
byhbt, line 121:
```suggestion
        Mix.raise(~s[Failed to create .keep file at #{path}. Reason: #{Atom.to_string(reason)}])
```

test/nimble.phx.gen.template/addons/ex_vcr_test.exs
```diff
@@ -0,0 +1,62 @@
+defmodule Nimble.Phx.Gen.Template.Addons.ExVCRTest do
+  use Nimble.Phx.Gen.Template.AddonCase
+
+  describe "#apply/2" do
+    @describetag mock_latest_package_versions: [{:exvcr, "0.12.2"}]
+
+    test "injects exvcr to mix dependency", %{
```
byhbt, line 7:
```suggestion
    test "injects ExVCR to mix dependency", %{
```

Update the case of `excvr` for the test description to make it easier to read.

andyduong1920:
Should we made this as an addon @hanam1ni because not all project need to use ExVCR 🤔

hanam1ni:
> Should we make this as an optianl addon @hanam1ni because not all project need to use ExVCR 🤔

@andyduong1920 Updated to make it optional
image

lib/nimble.phx.gen.template/generator.ex
hanam1ni: Updated the error message and use comma as a separator.

lib/nimble.phx.gen.template/addons/ex_vcr.ex
hanam1ni: For unused variables, we can name with underscore or prefix underscore with the name. In this project, we set Credo to always name unused variables with an underscore for clarification.

byhbt: thanks @hanam1ni !

olivierobert (approved):

byhbt (approved):

andyduong1920:
@hanam1ni The PR looks great since we made the ExVCR addon optional now so I think it better to test that addon on the integration test too, what do you think? we can do that by adding the `Y` answer in the Makefile https://github.com/nimblehq/elixir-templates/blob/develop/Makefile#L13

```
# Y - in response to Would you like to add the ExVCR addon?
common_addon_prompts = Y\nY\nY\nY\nY\n
```

andyduong1920:
Since we have merged the Mix variant, what do you think if we could add ExVCR addon into the Mix template too @hanam1ni 🙏

hanam1ni:
@andyduong1920 I updated the `Makefile` and refactored the modules, but I don't think we should add `ExVCR` in the mix template. 

To make the ExVCR work, we need to update the `config` and `cases` modules but the Mix project does not generate those by default. So I think we should not add `ExVCR` in the Mix template to make the project minimal.

andyduong1920:
> To make the ExVCR work, we need to update the `config` and `cases` modules but the Mix project does not generate those by default. So I think we should not add `ExVCR` in the Mix template to make the project minimal.

Yeah, agreed @hanam1ni

Issue #89, opened 2021-03-29
Adding `capture_log: true` to `ExUnit.start`


andyduong1920:
## Why

To not display the log info when the test is passed, in most cases when the test is passed, we don't need these logs.

https://elixirschool.com/blog/til-capture-log-in-exunit-tests/

PR #93, opened 2021-04-08
Adding capture_log: true

andyduong1920:
https://github.com/nimblehq/elixir-templates/issues/89

## What happened

Start ExUnit with `capture_log: true` option
 
## Insight

https://elixirschool.com/blog/til-capture-log-in-exunit-tests/
 
## Proof Of Work

The test is passed

### DIFF
--- a/.github/workflows/publish_to_hex_pm.yml
+++ b/.github/workflows/publish_to_hex_pm.yml
@@ -19,6 +19,11 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success' }}
     steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.8.0
+        with:
+          access_token: ${{ github.token }}
+      
       - uses: actions/checkout@v2
         with:
           ref: ${{ github.event.workflow_run.head_branch || github.ref }}
--- a/.github/workflows/test_api_variant_on_custom_api_project.yml
+++ b/.github/workflows/test_api_variant_on_custom_api_project.yml
@@ -24,6 +24,11 @@ jobs:
           --health-retries 5
         
     steps:
+      - name: Cancel Previous Runs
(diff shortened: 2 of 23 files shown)


hanam1ni:
How about we move the step that we update the `ExUnit.start` option after install all the addon? So we don't need to add the option in every addon and it will be easy to add other options in the future. Also, support when we want to update other modules.
```elixir
defp common_setup(%Project{} = project) do
    # Install addons ... 

    if install_addon_prompt?("Oban"), do: Addons.Oban.apply(project)
    if install_addon_prompt?("ExVCR"), do: Addons.ExVCR.apply(project)

    project
    |> update_test_helper()
    |> update_others() # In the future, we just pipe it.
end
```

andyduong1920:
> How about we move the step that we update the `ExUnit.start` option after install all the addon? So we don't need to add the option in every addon and it will be easy to add other options in the future. Also, support when we want to update other modules.
> 
> ```elixir
> defp common_setup(%Project{} = project) do
>     # Install addons ... 
> 
>     if install_addon_prompt?("Oban"), do: Addons.Oban.apply(project)
>     if install_addon_prompt?("ExVCR"), do: Addons.ExVCR.apply(project)
> 
>     project
>     |> update_test_helper()
>     |> update_others() # In the future, we just pipe it.
> end
> ```

Do you mean moving on the updating `ExUnit.start()` step or the whole test_env addon at the end @hanam1ni ?

If we move only the updating `ExUnit.start()` step, what should we name for that addon 🤔  ExUnit addon 🤔

hanam1ni (approved):

andyduong1920:
Oh I just merged https://github.com/nimblehq/elixir-templates/pull/95 to this branch 🙈. cc @byhbt

andyduong1920:
> Oh I just merged #95 to this branch 🙈. cc @byhbt

@byhbt for the ExUnit addon, we just need to review 5 files, the other files came from #95 :( 

image

byhbt (approved):

PR #95, opened 2021-04-09
Refactor the Github action flows

andyduong1920:
## What happened

Improving the GitHub Action test flow by

1/ Adding **Cancel Previous Runs**

2/ Refactor the **Cache Node npm**

3/ Separate the **mix test** and **mix compile** step
 
## Proof Of Work

The test is green

### DIFF
--- a/.github/workflows/publish_to_hex_pm.yml
+++ b/.github/workflows/publish_to_hex_pm.yml
@@ -19,6 +19,11 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success' }}
     steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.8.0
+        with:
+          access_token: ${{ github.token }}
+      
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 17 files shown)


byhbt:
hi @andyduong1920 !

btw, this PR is about improve the Github Action workflow, should we add the rule to ignore running the build for the Readme file update? :D 

Reference:
- https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
- https://github.community/t/using-on-push-tags-ignore-and-paths-ignore-together/16931

andyduong1920:
> hi @andyduong1920 !
> 
> btw, this PR is about improve the Github Action workflow, should we add the rule to ignore running the build for the Readme file update? :D
> 
> Reference:
> 
> * https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
> * https://github.community/t/using-on-push-tags-ignore-and-paths-ignore-together/16931

Hi @byhbt on some projects, it required all CI flows are passed in order to merge that PR, it doesn't matter what contact in that PR, all tests should be 🟢 , so if we skip on that file, the PR can't be merge @byhbt, so in general that should be a template, instead each project could define it by themself, what do you think @byhbt ?

hanam1ni (approved):

byhbt (approved):

PR #110, opened 2021-05-26
Enable the test release version workflow

andyduong1920:
## What happened

Enable the test Release version workflow
 
## Insight

Because we changed the package name to `nimble_template` on the previous version, at that time the `nimble_template` didn't exist on Hex yet so we need to disable this workflow on that release.

### DIFF
--- a/.github/workflows/test_release_version.yml
+++ b/.github/workflows/test_release_version.yml
@@ -1,53 +1,53 @@
-# name: "Test Release version"
+name: "Test Release version"
 
-# on:
-#   push:
-#     branches:
-#       - 'release/**'
+on:
+  push:
+    branches:
+      - 'release/**'
 
-# env:
-#   OTP_VERSION: 23.3
-#   ELIXIR_VERSION: 1.11.4
-#   PHOENIX_VERSION: 1.5.7
-#   MIX_ENV: test
+env:
+  OTP_VERSION: 23.3
+  ELIXIR_VERSION: 1.11.4
+  PHOENIX_VERSION: 1.5.7
+  MIX_ENV: test
 
-# jobs:
-#   release_version_test:
-#     runs-on: ubuntu-latest
+jobs:
+  release_version_test:
+    runs-on: ubuntu-latest
 
-#     steps:
-#       - name: Cancel Previous Runs
-#         uses: styfle/cancel-workflow-action@0.8.0
-#         with:
-#           access_token: ${{ github.token }}
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.8.0
+        with:
+          access_token: ${{ github.token }}
 
-#       - uses: actions/checkout@v2
-#         with:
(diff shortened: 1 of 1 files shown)


hanam1ni (approved):

.github/workflows/test_release_version.yml
```diff
@@ -1,53 +1,53 @@
...
+  PHOENIX_VERSION: 1.5.7
+  MIX_ENV: test
+
+jobs:
+  release_version_test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.8.0
```
longnd, line 20: The latest version is [0.9.0](https://github.com/styfle/cancel-workflow-action/releases). We can consider using it.

.github/workflows/test_release_version.yml
```diff
@@ -1,53 +1,53 @@
...
+        with:
+          access_token: ${{ github.token }}
+
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+
+     - uses: erlef/setup-elixir@v1
```
longnd, line 28: How about adding a name for the steps?
```suggestion
      - name: Checkout repository
        uses: actions/checkout@v2
        with:
          ref: ${{ github.head_ref }}

      - name: Setup Elixir
        uses: erlef/setup-elixir@v1
```

.github/workflows/test_release_version.yml
```diff
@@ -1,53 +1,53 @@
...
-#           path: |
-#             _build
-#             deps
-#           key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
-#           restore-keys: |
-#             ${{ runner.os }}-mix-
-
-#       - name: Install Dependencies
-#         run: mix deps.get
+name: "Test Release version"
```
longnd, line 1: The `"` character here isn't required. I suggest removing it to keep the consistency because we don't use it in the names of the steps
```suggestion
name: Test Release version
```

.github/workflows/test_release_version.yml
andyduong1920: I think it better to upgrade on another PR along with other workflow as well @longnd

.github/workflows/test_release_version.yml
andyduong1920: Adjusted in https://github.com/nimblehq/elixir-templates/pull/110/commits/1520f628b213a6613962dcc8eff58338da808cb0 thanks @longnd

.github/workflows/test_release_version.yml
andyduong1920: Adjusted in https://github.com/nimblehq/elixir-templates/pull/110/commits/1520f628b213a6613962dcc8eff58338da808cb0 thanks @longnd

longnd (approved):

andyduong1920:
Thanks @longnd @hanam1ni

PR #115, opened 2021-06-15
Upgrade Elixir version to 1.12.1

olivierobert:
## What happened

- Update Elixir to 1.12.2
- Update to Erlang/OTP to 24.0.4
 
## Insight

When building for production with the Dockerfile, the app crashed on Heroku 😱 Upon checking the log, it appeared that two C libraries were missing so I updated the Dockerfile. 

image

I tested this update on https://github.com/nimblehq/centauri-web/pull/115.

 
## Proof Of Work

Technically the fact that https://github.com/nimblehq/centauri-web is deployed onto Heroku is the proof of work 👉 https://centauri-staging.nimblehq.co/

### DIFF
--- a/.github/workflows/publish_to_hex_pm.yml
+++ b/.github/workflows/publish_to_hex_pm.yml
@@ -11,8 +11,8 @@ on:
   workflow_dispatch:
 
 env:
-  OTP_VERSION: 24.0.1
-  ELIXIR_VERSION: 1.12.1
+  OTP_VERSION: 24.0.4
+  ELIXIR_VERSION: 1.12.2
 
 jobs:
   publish:
--- a/.github/workflows/publish_to_hex_pm.yml
+++ b/.github/workflows/publish_to_hex_pm.yml
@@ -11,10 +11,10 @@ on:
   workflow_dispatch:
 
 env:
-  OTP_VERSION: 23.3
-  ELIXIR_VERSION: 1.11.4
+  OTP_VERSION: 24.0.1
+  ELIXIR_VERSION: 1.12.1
 
-jobs:   
+jobs:
   publish:
     runs-on: ubuntu-latest
     if: ${{ github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success' }}
@@ -23,7 +23,7 @@ jobs:
         uses: styfle/cancel-workflow-action@0.8.0
         with:
           access_token: ${{ github.token }}
-      
+
       - uses: actions/checkout@v2
         with:
           ref: ${{ github.event.workflow_run.head_branch || github.ref }}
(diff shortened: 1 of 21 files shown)


olivierobert:
@andyduong1920 The tests are failing due to test coverage issues:

image

So I recommend excluding these two built-in files from the test coverage:

```
{
  "skip_files": [
    ...
    "lib/centauri_web/telemetry.ex",
    "lib/centauri_web/channels/user_socket.ex",
    ...
  ],
  "coverage_options": {
    "minimum_coverage": 100
  }
}
```

What do you think?

olivierobert:
I went ahead with doing the required fixes to have all the tests passing ✅

andyduong1920:
Could you take a look at the `--warnings-as-errors` setup as on 1.12+ we don't need to setup some stuff as manually  @olivierobert 

https://dashbit.co/blog/tests-with-warnings-as-errors

image

andyduong1920:
Small thing, how about update the Elixir and Erlang to 

- Elixir: 1.12.2
- Erlang: 24.0.4

olivierobert:
@andyduong1920 I updated the more recent version of Elixir and Erlang/OTP in daa237c

However, regarding the removal of `--warnings-as-errors`, while it seems nice at first (also thanks for sharing that 🙏 ), I found that it would be less nice in terms of developer experience. For instance, in local development, we would need to type `mix test --warnings-as-errors` if we don't use `mix coverage` or if we don't add our own alias for `mix test` in `mix.exs`. I am concerned that some devs might not append `--warnings-as-errors` all the time locally (e.g. when using `mix test --failed`) causing them to only see the warnings on the CI which is too late 💭

andyduong1920:
> @andyduong1920 I updated the more recent version of Elixir and Erlang/OTP in [daa237c](https://github.com/nimblehq/elixir-templates/commit/daa237c9df7f17ae1f1143a2a9e28f0df200ea6a)
> 
> However, regarding the removal of `--warnings-as-errors`, while it seems nice at first (also thanks for sharing that 🙏 ), I found that it would be less nice in terms of developer experience. For instance, in local development, we would need to type `mix test --warnings-as-errors` if we don't use `mix coverage` or if we don't add our own alias for `mix test` in `mix.exs`. I am concerned that some devs might not append `--warnings-as-errors` all the time locally (e.g. when using `mix test --failed`) causing them to only see the warnings on the CI which is too late 💭

Ah true, we have to add `--warnings-as-errors` manually in order to have that, then I think we can keep it as default as its 😂

andyduong1920 (approved):

longnd (approved):

junan (approved):

byhbt (approved):

PR #123, opened 2021-09-17
Order mix dependencies

hanam1ni:
## What happened

Order dependencies in `mix.exs` in alphabetical order
 
## Proof Of Work

Setup project with the template, all dependencies are properly ordered.
image

### DIFF
--- a/lib/nimble_template/helpers/dependency.ex
+++ b/lib/nimble_template/helpers/dependency.ex
@@ -1,5 +1,34 @@
 defmodule NimbleTemplate.DependencyHelper do
-  def fetch_and_install_dependencies(),
+  alias NimbleTemplate.Generator
+
+  def fetch_and_install_dependencies,
     do:
       if(Mix.shell().yes?("\nFetch and install dependencies?"), do: Mix.shell().cmd("mix deps.get"))
+
+  def order_dependencies! do
+    file_content = File.read!("mix.exs")
+
+    dependencies = extract_dependencies(file_content)
+
+    ordered_dependencies =
+      dependencies
+      |> String.split(",\n")
+      |> Enum.sort()
+      |> Enum.join(",\n")
+
+    Generator.replace_content("mix.exs", dependencies, ordered_dependencies)
+  end
+
+  def extract_dependencies(contents) do
+    [_, deps_with_tail] =
+      :binary.split(contents, """
+        defp deps do
+          [
+      """)
+
+    [deps, _tail] =
+      :binary.split(deps_with_tail, "\n    ]\n  end")
+
+    deps
+  end
 end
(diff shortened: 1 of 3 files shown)


lib/nimble_template/helpers/dependency.ex
```diff
@@ -1,5 +1,34 @@
...
+      dependencies
+      |> String.split(",\n")
+      |> Enum.sort()
+      |> Enum.join(",\n")
+
+    Generator.replace_content("mix.exs", dependencies, ordered_dependencies)
+  end
+
+  def extract_dependencies(contents) do
+    [_, deps_with_tail] =
```
olivierobert, line 23: In terms of naming, while `tail` sounds idiomatic, `deps_with_tail` is not that descriptive as it requires reading the whole code to understand that the tail means the `end` block for the method.

olivierobert (commented):
Overall, it looks good to me. In terms of naming, it can be a bit ambiguous but I do not have a better idea. 

Note that the branch is failing for many variants.

lib/nimble_template/helpers/dependency.ex
hanam1ni: Updated the name and code formatting. So more variants of the test are passing now ✅

I think we should open another PR to fix the failing test variants

andyduong1920:
Please rebase this PR @hanam1ni the test was fixed on #128 👍

andyduong1920:
@hanam1ni should we add some test for this step 🤔

hanam1ni:
@andyduong1920 Added test in 6400316

test/nimble_template/helpers/dependency_test.exs
```diff
@@ -0,0 +1,32 @@
+defmodule NimbleTemplate.DependencyTest do
+  use NimbleTemplate.AddonCase, async: false
+
+  alias NimbleTemplate.{Addons, DependencyHelper}
+
+  describe "order_dependencies!/0" do
+    @describetag mock_latest_package_versions: [{:exvcr, "0.12.2"}]
+    @describetag mock_latest_package_versions: [{:mimic, "1.3.1"}]
```
andyduong1920, line 8: @hanam1ni can we try 

```
@describetag mock_latest_package_versions: [{:exvcr, "0.12.2"}, {:mimic, "1.3.1"}]
```

lib/nimble_template/helpers/dependency.ex
```diff
@@ -1,5 +1,33 @@
...
+      |> String.split(",\n")
+      |> Enum.sort()
+      |> Enum.join(",\n")
+
+    Generator.replace_content("mix.exs", dependencies, ordered_dependencies)
+  end
+
+  defp extract_dependencies(contents) do
+    [_, deps_with_file_footer] =
+      :binary.split(contents, """
```
andyduong1920, line 24: @hanam1ni I'm curious can we use String module instead of using the erlang module 🤔

```
String.split

vs

:binary.split
```

andyduong1920 (approved):

byhbt (approved):

PR #129, opened 2021-12-07
Skip node dependencies installation for API variant

hanam1ni:
## What happened

- CI is failing when running setup on API variants because these variants do not have `package.json`
image
 
- Solve the issue by installing node dependencies only when `package.json` exists.

## Proof Of Work

CI passed ✅

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -35,13 +35,10 @@ runs:
         MIX_ENV: test
 
     - name: Install Node Dependencies
-      run: |
-            cd sample_project 
-            if [ -f assets/package.json ]; then
-              npm install --prefix assets
-            else 
-              echo package.json does not exist, skipping node dependencies installation.
-            fi
+      env: 
+        VARIANT_WITHOUT_NODE: '["api"]'
+      if: contains(fromJson(env.VARIANT_WITHOUT_NODE), inputs.variant) != true
+      run: cd sample_project && npm install --prefix assets
       shell: bash
 
     - name: Run mix ecto.create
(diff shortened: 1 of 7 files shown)


andyduong1920:
I'm wondering why does the CI pass on https://github.com/nimblehq/elixir-templates/pull/128 🤔 is it something changed in the node, npm version which on the previous build they don't strict in the `packge.json` file while on the new node, npm version, they make it strickly no if there is no `package.json` file, they will raise the error 🤔

hanam1ni:
@andyduong1920 I am not sure about this one because I checked output in #128 CI, it displayed only warnings and did not stop the pipeline.
image

andyduong1920:
> @andyduong1920 I am not sure about this one because I checked output in #128 CI, it displayed only warnings and did not stop the pipeline. image

Yeah found that too, 
image

Because on that npm version, it allows running `npm install` without package.json file while on this new version, it doesn't

So the solution is doesn't run that npm install for API variant tests

action.yml
```diff
@@ -33,8 +33,11 @@ runs:
       shell: bash
       env:
         MIX_ENV: test
-      
+
     - name: Install Node Dependencies
+      env: 
+        VARIANT_WITHOUT_NODE: '["api"]'
```
andyduong1920, line 39: Hi @hanam1ni should we revert this ENV a bit to `VARIANT_WITH_NODE` to make it easy to understand 🤔 IMO or we could change to another name like

```
VARIANT_WITH_NODE_DEPENDENCY: '["web", "live"]'

if: contains(fromJson(env.VARIANT_WITH_NODE_DEPENDENCY), inputs.variant) == true
```

andyduong1920 (approved):

olivierobert (approved):

Issue #127, opened 2021-11-30
Add Base setup for API projects


andyduong1920:
## Why

For the API project, we need to set up some Plug, Param Validation, or Error View, we could inherit these configurations from our existing API projects. 

## Plan
- [x] Request Params Validation.
- [x] ErrorView return follow [JSON API](https://jsonapi.org/format) standard format. 
- [ ] Add fallback controller. (add `action_fallback MODULE.Api.V1.FallbackController` into web entry point)
- [ ] Check empty request body params Plug.
- [ ] Add `jsonapi` dependency to `mix.exs`.
- [ ] Add `alias MODULE.ParamsValidator` into web entry point.  

## Who Benefits?

The new API project's developer 🤞

andyduong1920:
Hi @byhbt can you share a plan on which component will you add into the API variant (make it as a TODO list) on this issue so we can keep track easily 

- [ ] Param Validation
- [ ] API error view
- [ ] Fallback controller
- [ ] ....

andyduong1920:
Hi @byhbt 👋did you check the comment above :)

PR #132, opened 2021-12-21
Add Param Validation for the API project

byhbt:
https://github.com/nimblehq/elixir-templates/issues/127

## What happened

- Implement setup Param validation for API projects. I will implement add error view in another PR.
 
⚠️  Notice: 
in this pull request, I renamed the `CustomModule` to `SampleCustomModule` and `custom_app` to `sample_custom_app` to make the name of the module consistent alphabetically with `sample_project`. In case we add a pre-defined alias module we will not have an issue with the Credo check for sorting the alias alphabetically warning.

Screen Shot 2021-12-22 at 11 00 57 AM

## Insight

- [x] Add param validations
- [x] Update the `.credo.exs` config of the output template.
  - [x] Allow `TODO` comment with exit code zero
  - [x] Disable check duplicated code in `tests`

## Proof Of Work

The test should pass.

### DIFF
--- a/priv/templates/nimble_template/.credo.exs
+++ b/priv/templates/nimble_template/.credo.exs
@@ -87,7 +87,7 @@
         # If you don't want TODO comments to cause `mix credo` to fail, just
         # set this value to 0 (zero).
         #
-        {Credo.Check.Design.TagTODO, [exit_status: 2]},
+        {Credo.Check.Design.TagTODO, [exit_status: 0]},
         {Credo.Check.Design.TagFIXME, []},
 
         #
@@ -161,9 +161,9 @@
            order:
              ~w/shortdoc moduledoc behaviour use import alias require module_attribute defstruct callback/a
          ]},
-        {Credo.Check.Consistency.MultiAliasImportRequireUse, []},
+        {Credo.Check.Consistency.MultiAliasImportRequireUse, false},
         {Credo.Check.Consistency.UnusedVariableNames, false},
-        {Credo.Check.Design.DuplicatedCode, []},
+        {Credo.Check.Design.DuplicatedCode, files: %{excluded: ["**/*_test.exs"]}},
         {Credo.Check.Readability.AliasAs, false},
         {Credo.Check.Readability.MultiAlias, false},
         {Credo.Check.Readability.Specs, false},
--- a/priv/templates/nimble_template/.credo.exs
+++ b/priv/templates/nimble_template/.credo.exs
@@ -161,7 +161,7 @@
            order:
              ~w/shortdoc moduledoc behaviour use import alias require module_attribute defstruct callback/a
          ]},
-        {Credo.Check.Consistency.MultiAliasImportRequireUse, false},
+        {Credo.Check.Consistency.MultiAliasImportRequireUse, []},
         {Credo.Check.Consistency.UnusedVariableNames, false},
         {Credo.Check.Design.DuplicatedCode, files: %{excluded: ["**/*_test.exs"]}},
         {Credo.Check.Readability.AliasAs, false},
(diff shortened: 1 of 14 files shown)


priv/templates/nimble_template/.credo.exs
```diff
@@ -161,9 +161,9 @@
            order:
              ~w/shortdoc moduledoc behaviour use import alias require module_attribute defstruct callback/a
          ]},
-        {Credo.Check.Consistency.MultiAliasImportRequireUse, []},
+        {Credo.Check.Consistency.MultiAliasImportRequireUse, false},
```
andyduong1920, line 164: is there any reason to disable this @byhbt

priv/templates/nimble_template/lib/otp_app_web/params/params.ex.eex
```diff
@@ -0,0 +1,23 @@
+defmodule <%= base_module %>Web.Params do
```
andyduong1920, line 1: can we try `<%= web_module %>` instead of `<%= base_module %>Web`

same for other places

andyduong1920 (commented):
Should we remove this file @byhbt 

image

priv/templates/nimble_template/.credo.exs
byhbt: Because as our compass, we allow import multiple alias 

Screen Shot 2021-12-23 at 10 51 39 AM

Reference: https://nimblehq.co/compass/development/code-conventions/elixir/#syntax

andyduong1920: That isn't correct @byhbt I believe we missed the other options (import, require, use) in our compass as we created that before we applied credo.

For me, better to follow credo/elixir convention @byhbt so let not disable it @byhbt

andyduong1920: but I do not fully understand you @byhbt what do you mean by `we allow import multiple alias` 🤔 ---> that is the default config for `MultiAliasImportRequireUse` 🤔, it works as expected, have you research what is the `MultiAliasImportRequireUse` rule in Credo before `disable` it

byhbt: Here is the example, when i tried to follow the compass convention for importing multiple aliases with the same parent namespace. 

Screen Shot 2021-12-23 at 11 54 30 AM

Reference: https://hexdocs.pm/credo/Credo.Check.Consistency.MultiAliasImportRequireUse.html

If we don't disable it, then we cannot apply the compass convention for this case.

andyduong1920: so that means there is some other place that is using `multiple alias` right @byhbt should we adjust it instead of `disable the credo` 🤔

As I understand, by disabling this credo rule here, in the long term we will have a mix between `multiple aliases` and single alias in our codebase which won't be good.

priv/templates/nimble_template/lib/otp_app_web/params/params.ex.eex
byhbt: Updated in https://github.com/nimblehq/elixir-templates/pull/132/commits/24b9962d27018316fa95caa8e149e1d269098320

priv/templates/nimble_template/.credo.exs
byhbt: After discuss via Huddle we agree to keep the credo check for `MultiAliasImportRequireUse`. https://github.com/nimblehq/elixir-templates/pull/132/commits/089f5c7e5e6823c8940d5db2a531e34e7ff39db7

rosle (approved):

andyduong1920 (approved):

Issue #107, opened 2021-05-18
As a developer, I should be noticed if I forgot to update localization files (PO, POT files)


andyduong1920:
## Why

Making sure the localization are up-to-date on each Pull request

## Who Benefits?

Developer

andyduong1920:
Reopen this issue as the implementation on https://github.com/nimblehq/elixir-templates/pull/138 doesn't work in case the 

- File localization file isn't created --> new file, git diff is empty
- The file is generated but the ci still passed as it's missing the `exit 1` to terminate the signal.

andyduong1920:
Revert PR: https://github.com/nimblehq/elixir-templates/pull/174

andyduong1920:
We have the native support from Gettext https://github.com/nimblehq/elixir-templates/pull/244

PR #138, opened 2021-12-23
[107] As a developer, I should be noticed if I forgot to update localization files (PO, POT files)

byhbt:
https://github.com/nimblehq/elixir-templates/issues/107

## What happened

As a developer, I should be noticed if I forgot to update localization files (PO, POT files)
 
## Insight

- [x] Add `alias` `gettext.extract-and-merge` for extract the translation phrases.
- [x] Add `alias` `gettext.check`: for checking if there is any missing extracted phrase from files.
  - `git diff --no-ext-diff --quiet priv\/gettext`. Reference: https://git-scm.com/docs/git-diff
    - `--no-ext-diff`: Disallow external diff drivers.
    - `--quiet`: Disable all output of the program. Implies `--exit-code`.
- [x] Combine the translation check into the `mix codebase` command. 

## Proof Of Work

The test should pass.

Output where there is some translation update in the codebase which is not being committed. 

Screen Shot 2022-01-10 at 2 29 15 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,7 @@ create_mix_project:
 # Y - in response to Do you want to publish a Github Wiki for this project? You'd need to manually create the first Github Wiki Page and set the GH_TOKEN and GH_EMAIL secret for this to properly function.
 # Y - in response to Would you like to add the Oban addon?
 # Y - in response to Would you like to add the ExVCR addon?
-# Y - in response to Would you like to add the Gettext Extract and Merge command addon?
-common_addon_prompts = Y\nY\nY\nY\nY\nY\nY\nY\nY\n
+common_addon_prompts = Y\nY\nY\nY\nY\nY\nY\nY\n
 
 # Y - in response to Would you like to add the SVG Sprite addon?
 # Y - in response to Would you like to add the Nimble CSS addon?
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,8 @@ create_mix_project:
 # Y - in response to Do you want to publish a Github Wiki for this project? You'd need to manually create the first Github Wiki Page and set the GH_TOKEN and GH_EMAIL secret for this to properly function.
 # Y - in response to Would you like to add the Oban addon?
 # Y - in response to Would you like to add the ExVCR addon?
-common_addon_prompts = Y\nY\nY\nY\nY\nY\nY\nY\n
+# Y - in response to Would you like to add the Gettext Extract and Merge command addon?
+common_addon_prompts = Y\nY\nY\nY\nY\nY\nY\nY\nY\n
 
 # Y - in response to Would you like to add the SVG Sprite addon?
 # Y - in response to Would you like to add the Nimble CSS addon?
@@ -27,8 +28,7 @@ web_addon_prompts = Y\nY\nY\n
 
 api_addon_prompts =
 
-
-live_addon_prompts = 
+live_addon_prompts =
 
 # Y - in response to Will you host this project on Github?
 # Y - in response to Do you want to generate the .github/ISSUE_TEMPLATE and .github/PULL_REQUEST_TEMPLATE?
--- a/lib/nimble_template/addons/gettext.ex
+++ b/lib/nimble_template/addons/gettext.ex
@@ -1,14 +1,14 @@
 defmodule NimbleTemplate.Addons.Gettext do
   @moduledoc false
 
-  use NimbleTemplate.Addon
+  use NimbleTemplate.Addons.Addon
 
   @impl true
   def do_apply(%Project{} = project, _opts) do
     edit_mix(project)
   end
 
-  defp edit_mix(%Project{mix_project?: true} = project) do
+  defp edit_mix(%Project{} = project) do
     Generator.inject_content(
       "mix.exs",
       """
--- a/lib/nimble_template/addons/gettext.ex
+++ b/lib/nimble_template/addons/gettext.ex
@@ -1,0 +1,25 @@
+defmodule NimbleTemplate.Addons.Gettext do
+  @moduledoc false
+
+  use NimbleTemplate.Addon
+
+  @impl true
+  def do_apply(%Project{} = project, _opts) do
+    edit_mix(project)
+  end
+
+  defp edit_mix(%Project{mix_project?: true} = project) do
+    Generator.inject_content(
+      "mix.exs",
+      """
+        defp aliases do
+          [
+      """,
+      """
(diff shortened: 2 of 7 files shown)


lib/nimble_template/addons/gettext.ex
```diff
@@ -0,0 +1,25 @@
...
+
+  defp edit_mix(%Project{mix_project?: true} = project) do
+    Generator.inject_content(
+      "mix.exs",
+      """
+        defp aliases do
+          [
+      """,
+      """
+        "gettext.extract-and-merge": ["gettext.extract --merge --no-fuzzy"],
```
andyduong1920, line 19: should we add `gettext.extract-and-merge` to the `codebase.fix` alias @byhbt 🤔

priv/templates/nimble_template/.github/workflows/test.yml.eex
```diff
@@ -88,6 +88,15 @@ jobs:
       - name: Run codebase check
         run: mix codebase
 
+      - name: Ensure that localization files (POs, POTs) are up-to-date.
+        run:
+          mix gettext.extract-and-merge;
+
+          if [ -n "$(git diff --exit-code priv/gettext/)" ]; then
+            echo "The localization files (POs, POTs) are NOT up-to-date, run \"mix gettext.extract-and-merge\" on your local and push again.";
```
andyduong1920, line 96: can we have this condition in the `mix codebase` @byhbt so the developer can check it on local 🤔

also for the message, if we add the `ettext.extract-and-merge` into `codebase.fix` we can adjust the message here to `The localization files (POs, POTs) are NOT up-to-date, run \"mix codebase.fix`

byhbt: hi @andyduong1920 

I have one problem with this, because the `codebase.fix` command is injected via `NimbleTemplate.Addons.TestEnv`, if we include the `gettext.extract-and-merge` command to `codebase.fix`, then this add-on will depend on `TestEnv`. 

So if we want to add `gettext.extract-and-merge` to `codebase.fix` we need to accept the depend on `TestEnv`, to make sure that the `codebase.fix` is existed before we inject this command. 

Here is the screenshot of `mix.exs` content after apply `Gettext` addon only.

Screen Shot 2022-01-06 at 2 53 01 PM

andyduong1920: Hi @byhbt you can check the `@describetag required_addons: [:TestEnv]` in the test codebase :)

lib/nimble_template/addons/gettext.ex
byhbt: Added in https://github.com/nimblehq/elixir-templates/pull/138/commits/eeb9256c0af1ff17907c8aee83b76a1e67a2ca6d

lib/nimble_template/templates/variants/phoenix/template.ex
```diff
@@ -67,6 +67,9 @@ defmodule NimbleTemplate.Templates.Phoenix.Template do
     if install_addon_prompt?("Oban"), do: PhoenixAddons.Oban.apply(project)
     if install_addon_prompt?("ExVCR"), do: PhoenixAddons.ExVCR.apply(project)
 
+    if install_addon_prompt?("Gettext Extract and Merge command"),
```
andyduong1920, line 70: should we name it as `Gettext Helpers` @byhbt 🤔 

`if install_addon_prompt?("Gettext Helpers"),`

test/nimble_template/addons/variants/gettext_test.exs
```diff
@@ -0,0 +1,44 @@
...
+        assert_file("mix.exs", fn file ->
+          assert file =~ """
+                      "codebase.fix": [
+                         "gettext.extract-and-merge",
+                 """
+        end)
+
+        assert_file("mix.exs", fn file ->
+          assert file =~ """
+                       codebase: [\n    \"gettext.check\",
```
andyduong1920, line 38: can we use

```
codebsae: [
  "gettext.check"
```

priv/templates/nimble_template/.github/workflows/test.yml.eex
```diff
@@ -88,6 +88,15 @@ jobs:
       - name: Run codebase check
         run: mix codebase
 
+      - name: Ensure that localization files (POs, POTs) are up-to-date.
```
andyduong1920, line 91: @byhbt do we need this step as we have the `gettext.check` in the `mix codebase` above 🤔

Makefile
```diff
@@ -18,7 +18,8 @@ create_mix_project:
 # Y - in response to Do you want to publish a Github Wiki for this project? You'd need to manually create the first Github Wiki Page and set the GH_TOKEN and GH_EMAIL secret for this to properly function.
 # Y - in response to Would you like to add the Oban addon?
 # Y - in response to Would you like to add the ExVCR addon?
-common_addon_prompts = Y\nY\nY\nY\nY\nY\nY\nY\n
+# Y - in response to Would you like to add the Gettext Extract and Merge command addon?
```
olivierobert, line 21: Why not make this added by default i.e., so not an addon? All projects need to be localized and even if there is only one language, the tasks must be there.

byhbt: Yup, that's also a good idea, as i see the default generated Phoenix project is included the Gettext and use it from the start.

Screen Shot 2022-02-18 at 3 28 20 PM

test/nimble_template/addons/variants/gettext_test.exs
byhbt: Updated in https://github.com/nimblehq/elixir-templates/pull/138/commits/1f89b22394274effb350dded98dbcb46fe0472db

priv/templates/nimble_template/.github/workflows/test.yml.eex
byhbt: right, it's duplicated, i removed it https://github.com/nimblehq/elixir-templates/pull/138/commits/139ab3f040af54805e54be98656be3b0e07dc371

lib/nimble_template/templates/variants/phoenix/template.ex
byhbt: yup, as [discussed](https://github.com/nimblehq/elixir-templates/pull/138#discussion_r809772565) with Olivier, we agreed that we can make this addon as default because initially the generated Phoenix project already included the `Gettext`, so I will remove this `prompt` also.

andyduong1920 (approved):

olivierobert (approved):

PR #101, opened 2021-05-12
Add the workflow to deploy to Heroku and release PR template

longnd:
## What happened

- [x] Add a workflow to deploy to Heroku
- [x] Add the template for the Release PR to follow our [convention](https://nimblehq.co/compass/development/version-control/release-management/#release-pull-request)
 
## Insight

- Inspired by [this article](https://dev.to/heroku/deploying-to-heroku-from-github-actions-29ej) I add a Github Action workflow to deploy to Heroku using Container stack instead of buildpacks. I think it's cleaner since there are fewer files to manage. The deployment scripts for Heroku are located in the Github workflow files only, no Procfile & buildpack config files are needed.

- Also add the REAMDE file with usage instruction that requires some configuration changes, following the [official documentation](https://hexdocs.pm/phoenix/heroku.html#making-our-project-ready-for-heroku)

- This PR also resolve an issue of conflict dependency version in the web variant:
  - When initializing a new Phoenix project, internally it depends on **Plug** dependency which also has **Mime**("~> 1.0 or ~> 2.0") a dependency. As a result, mime v2 is used and defined in `mix.lock`
  -  When apply this elixir-template to the project to generate the web variant, it uses `Wallaby` as a dependency. Wallaby itself relies on **Telsa** -> **mime** (~1.0) as dependency.
  - Due to this **mime** versions conflict, running `mix deps.get` will not work, we must resolve the conflict by allowing **Tesla** to override the existing **mime** version to uses its version.
    ```elixir
    {:tesla, latest_package_version(:tesla), override: true}
    ```
 
## Proof Of Work
Initialize a new project using this Elixir template:
Screen Shot 2021-08-15 at 11 23 52

I used this workflow in my IC - implemented in [this PR](https://github.com/longnd/elixir-gscraper/pull/27)
Screen Shot 2021-04-09 at 14 16 25

PR #141, opened 2021-12-28
Generate a corresponding .github/README content

andyduong1920:
## What happened

To solve https://github.com/nimblehq/elixir-templates/pull/101/files#r775813945

## Insight

1/ Add the condition into the Readme file.
2/ Pass the condition from 

image

3/ Grab the condition value from

image

## Proof Of Work

The test is passed 💙


lib/nimble_template/addons/variants/phoenix/web/wallaby.ex
```diff
@@ -42,10 +42,13 @@ defmodule NimbleTemplate.Addons.Phoenix.Web.Wallaby do
...
   defp inject_mix_dependency(%Project{} = project) do
-    Generator.inject_mix_dependency([
+    Generator.inject_mix_dependency(
       {:wallaby, latest_package_version(:wallaby), only: :test, runtime: false}
-    ])
+    )
 
+    # There is a conflict on `mime` version between Plug and Tesla.
+    # The Wallaby team hasn't used a the new Tesla version 1.4.4.
+    # TODO: Remove the mime update once Wallaby team uses the new Tesla version.
```
longnd, line 51: Since Wallaby use `web_driver_version` as dependency
https://github.com/elixir-wallaby/web_driver_client/blob/master/mix.exs#L42
> {:web_driver_client, "~> 0.2.0"},

and `web_driver_version` use `testla`
https://github.com/elixir-wallaby/web_driver_client/blob/8dc46a325f90a120cfa3c65c19da68faef553299/mix.exs#L42
> {:tesla, "~> 1.3"},

so I think testla 1.4 should match right? 🤔

andyduong1920: Oh good catch @longnd let me check that

andyduong1920: yeah it works @longnd I have removed that in https://github.com/nimblehq/elixir-templates/pull/141/commits/b5b338ac3c07869d7bc0aaef25550517cce6611b

byhbt (approved):
LGTM 🚀

priv/templates/nimble_template/.github/workflows/README.md.eex
```diff
@@ -1,13 +1,14 @@
...
 
-The following workflows are supported. Depends on what you choose when generating the project, some or all of them would be available:
+The following workflows are supported.
 
+<%= if with_test_workflow? do %>
 - Test
-- [Deploy to Heroku](#deploy-to-heroku-workflow)
+<% end %>
+<%= if with_deploy_to_heroku_workflow? do %>
+- [Deploy to Heroku](#deploy-to-heroku-workflow-usage-instruction)
```
longnd, line 9: I think this part should be wrapped in its own `if` clause because this section is for the Worklow list 
so we can have more workflows added later
```suggestion
<%= if with_deploy_to_heroku_workflow? do %>
- [Deploy to Heroku](#deploy-to-heroku-workflow-usage-instruction)
<% end %>

# and later we can have
<%= if with_deploy_to_ecs_workflow? do %>
- [Deploy to ECS](#deploy-to-ECS-workflow-usage-instruction)
<% end %>
```

longnd (approved):
The rest LGTM

priv/templates/nimble_template/.github/workflows/README.md.eex
andyduong1920: Yeah @longnd I was thinking about that, but for now, I just follow what we have in the current Readme content also I'm not sure we will have both Heroku and ECS content, but anw I will separate it to make it easy to edit in the future

andyduong1920: Adjusted in https://github.com/nimblehq/elixir-templates/pull/141/commits/0fd1ca2c51a7b792855bce98b0a42b003714e8fa

Issue #44, opened 2020-11-12
Manual Ecto migrations addon


andyduong1920:
Following the https://dashbit.co/blog/automatic-and-manual-ecto-migrations it nice to have the `manual_migration` pre-setting

1/ Adding migrate_manual to the https://github.com/nimblehq/elixir-templates/blob/develop/priv/templates/nimble.phx.gen.template/lib/otp_app/release_tasks.ex.eex

```
  def migrate_manual do
    load_app()

    for repo <- repos() do
      run_migrations(repo, "priv/repo/manual_migrations")
    end
  end
```

2/ Adjust aliases in mix.ex file

```
defp aliases() do
  [
    "ecto.migrate_all": ["ecto.migrate --migrations-path=priv/repo/migrations --migrations-path=priv/repo/manual_migrations"],
    "ecto.setup": ["ecto.create", "ecto.migrate_all", "run priv/repo/seeds.exs"],
    test: ["ecto.create --quiet", "ecto.migrate_all --quiet", "test"],
    ...
  ]
end
```

andyduong1920:
@nimblehq/elixir-guild Do you think it worth to add this Addon into the template?

rosle:
@andyduong1920 I believe this now becomes the `data_migrations`, corrrect?

There is a specific config that we've added into one of the project. Because we're trying to avoid the data migration on the test.

With the config:

```elixir
defp aliases() do
  [
    "ecto.migrate_all": ["ecto.migrate --migrations-path=priv/repo/migrations --migrations-path=priv/repo/manual_migrations"],
    "ecto.setup": ["ecto.create", "ecto.migrate_all", "run priv/repo/seeds.exs"],
    test: ["ecto.create --quiet", "ecto.migrate --quiet", "test"]
```

This works perfectly, when running `mix test`, it is not running the data migration. However there is another problem when we run `MIX_ENV=test mix ecto.setup`, the data migration is run. 💡 

So we did:

```elixir
defp aliases() do
  [
    "ecto.setup": ["ecto.create", &migrate/1, "run priv/repo/seeds.exs"],
    test: ["ecto.create --quiet", &migrate/1, "test"],
    ...
  ]
end

defp migrate(_) do
  if Mix.env() == :test do
    Mix.Task.run("ecto.migrate", ["--quiet"])
  else
    Mix.Task.run("ecto.migrate", [
      "--migrations-path=priv/repo/migrations",
      "--migrations-path=priv/repo/data_migrations"
[...]

PR #148, opened 2022-01-04
Add data-migration

andyduong1920:
https://github.com/nimblehq/elixir-templates/issues/44#issuecomment-1004532089

## What happened

Add the `data_migration` as it's becoming the standard structure in Nimble projects

## Insight

1/ Create the `priv/repo/data_migrations/` directory
2/ Add `migrate_schemas` and `migrate_data` into the release task
3/ Add `mix ecto.migrate_all` task
4/ Adjust the `mix ecto.setup` 
- On Test ENV: does not run the `data migration`
- On Development: run both `schemas and data` migrations

## Proof Of Work

The test is passed

### DIFF
--- a/priv/templates/nimble_template/lib/otp_app/release_tasks.ex.eex
+++ b/priv/templates/nimble_template/lib/otp_app/release_tasks.ex.eex
@@ -4,8 +4,16 @@ defmodule <%= base_module %>.ReleaseTasks do
   def migrate do
     load_app()
 
-    migrate_schemas()
-    migrate_data()
+    for repo <- repos() do
+      schema_migrations = Ecto.Migrator.migrations_path(repo, "migrations")
+      data_migrations = Ecto.Migrator.migrations_path(repo, "data_migrations")
+
+      {:ok, _, _} =
+        Ecto.Migrator.with_repo(
+          repo,
+          &Ecto.Migrator.run(&1, [schema_migrations, data_migrations], :up, all: true)
+        )
+    end
   end
 
   def rollback(repo, version) do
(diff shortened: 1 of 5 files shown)


rosle (approved):

priv/templates/nimble_template/lib/otp_app/release_tasks.ex.eex
```diff
@@ -29,4 +28,18 @@ defmodule <%= base_module %>.ReleaseTasks do
...
+    for repo <- repos() do
+      {:ok, _, _} = Ecto.Migrator.with_repo(repo, &Ecto.Migrator.run(&1, :up, all: true))
+    end
+  end
+
+  defp migrate_data do
+    for repo <- repos() do
+      source = Ecto.Migrator.migrations_path(repo, "data_migrations")
+
+      {:ok, _, _} = Ecto.Migrator.with_repo(repo, &Ecto.Migrator.run(&1, source, :up, all: true))
```
bterone, line 42: Small question, can we pass a list as the sources for migrations? 

Then we could just have both the base `migrations path` and `data_migrations path` in a list like `&Ecto.Migrator.run(&1, [base_source, data_source], ...)`

and avoid [the issue](https://nimble-co.slack.com/archives/C0213CHBEPP/p1629191907014800?thread_ts=1629191470.013600&cid=C0213CHBEPP) that we previously had with the order of the migrations 🤔 .

andyduong1920: Hi, @bterone according to the Ecto.Migrator codebase (https://github.com/elixir-ecto/ecto_sql/blob/v3.7.1/lib/ecto/migrator.ex#L458) it seems like doable but we haven't tested that on Production 😂 let me check that a bit, thanks for the suggestion 👍

andyduong1920: Hi @bterone I can confirm it is working, I will adjust the template :)

image

andyduong1920: Update it in https://github.com/nimblehq/elixir-templates/pull/148/commits/eb672a75f78ed4163e42be6bef770de061715f2b @bterone

bterone (approved):

Issue #147, opened 2022-01-04
Mention `Github wiki` in Readme incase the project has Github Wiki documentation


andyduong1920:
## Why

Better to mention the `Github wiki` in the README so the developer is noticed that there is more documentation in Wiki

## Who Benefits?

developer

PR #149, opened 2022-01-04
Add `Wiki section` into README.md

andyduong1920:
https://github.com/nimblehq/elixir-templates/issues/147

## What happened

Mention Wiki in README.md

## Proof Of Work

The test is passed

image

### DIFF
--- a/lib/nimble_template/addons/github.ex
+++ b/lib/nimble_template/addons/github.ex
@@ -89,16 +89,22 @@ defmodule NimbleTemplate.Addons.Github do
   end
 
   @impl true
-  def do_apply(
-        %Project{
-          web_project?: web_project?,
-          mix_project?: mix_project?,
-          erlang_version: erlang_version,
-          elixir_version: elixir_version
-        } = project,
-        opts
-      )
-      when is_map_key(opts, :github_wiki) do
+  def do_apply(%Project{} = project, opts) when is_map_key(opts, :github_wiki) do
+    project
+    |> copy_wiki_files()
+    |> append_wiki_into_readme()
+
+    project
+  end
+
+  defp copy_wiki_files(
+         %Project{
+           web_project?: web_project?,
+           mix_project?: mix_project?,
+           erlang_version: erlang_version,
+           elixir_version: elixir_version
+         } = project
+       ) do
     binding = [
       web_project?: web_project?,
       mix_project?: mix_project?,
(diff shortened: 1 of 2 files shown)


lib/nimble_template/addons/github.ex
```diff
@@ -128,4 +134,18 @@ defmodule NimbleTemplate.Addons.Github do
...
   end
+
+  defp append_wiki_into_readme(project) do
+    Generator.append_content(
+      "README.md",
+      """
+
+      ## Project documentation
+
+      Most of the documentation locate in the `.github/wiki` directory, which is published to the [project's Github wiki](https://github.com/[REPO]/wiki).
```
bterone, line 145:
```suggestion
      Most of the documentation is located in the `.github/wiki` directory, which is published to the [project's Github wiki](https://github.com/[REPO]/wiki).
```

bterone (approved):
LGTM 😄

lib/nimble_template/addons/github.ex
```diff
@@ -89,16 +89,22 @@ defmodule NimbleTemplate.Addons.Github do
...
-        %Project{
-          web_project?: web_project?,
-          mix_project?: mix_project?,
-          erlang_version: erlang_version,
-          elixir_version: elixir_version
-        } = project,
-        opts
-      )
-      when is_map_key(opts, :github_wiki) do
+  def do_apply(%Project{} = project, opts) when is_map_key(opts, :github_wiki) do
```
rosle, line 92: Just wondering where does it fallback to when `github_wiki` does not present in the options? 💭

rosle (approved):

lib/nimble_template/addons/github.ex
andyduong1920: Thanks Terone, I have updated it in https://github.com/nimblehq/elixir-templates/pull/149/commits/a585cf3ceb677759e2807d77521d09fa37982aa4

lib/nimble_template/addons/github.ex
andyduong1920: Hi Ros, if we don't pass that key then it will look for another key

image

In case no key is passed into the `Github apply` then the exception will be raised.

In that case, there will be something wrong internally and we need to fix that :)

rosle: I see. Seems the opts is a map, so we also could do something like this


```suggestion
  def do_apply(%Project{} = project, %{github_wiki: true}) do
```
🤔 💡

andyduong1920: Hi @rosle I don't remember much the reason the team used `when is_map_key(opts, :github_wiki)` let me check the git log again :D

andyduong1920: I have opened a chore PR for that FYI @rosle https://github.com/nimblehq/elixir-templates/pull/152

PR #150, opened 2022-01-04
Enable Gzip for Static

andyduong1920:
https://github.com/nimblehq/elixir-templates/issues/104

## What happened

Enable `gzip: true` for static

## Insight

N/A

## Proof Of Work

The test is passed

### DIFF
--- a/lib/nimble_template/addons/variants/phoenix/web/assets.ex
+++ b/lib/nimble_template/addons/variants/phoenix/web/assets.ex
@@ -12,6 +12,7 @@ defmodule NimbleTemplate.Addons.Phoenix.Web.Assets do
     project
     |> edit_mix()
     |> edit_assets_package()
+    |> gzip_plug_static()
 
     project
   end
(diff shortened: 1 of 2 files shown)


bterone (approved):

lib/nimble_template/addons/variants/phoenix/web/assets.ex
```diff
@@ -61,4 +62,18 @@ defmodule NimbleTemplate.Addons.Phoenix.Web.Assets do
 
     project
   end
+
+  defp gzip_plug_static(%Project{web_path: web_path} = project) do
```
rosle, line 66: How about `enable_gzip_for_static_assets` 💡

rosle (approved):

lib/nimble_template/addons/variants/phoenix/web/assets.ex
andyduong1920: Thanks Ros, changed in https://github.com/nimblehq/elixir-templates/pull/150/commits/17556f1693ae7fdfa75be47ccea121823c0de6d9 :D

PR #152, opened 2022-01-04
Refactor the pattern matching in the Github addon

andyduong1920:
https://github.com/nimblehq/elixir-templates/pull/149#discussion_r777949010

## What happened

Minor refactor the pattern matching in Github addon

## Insight

We don't need to catch the `github_action_deploy_heroku: false` as from the Template module, we won't pass the `false` value, so if there is a case with `github_action_deploy_heroku: false` then better to let it raises an exception

## Proof Of Work

The test is passed

### DIFF
--- a/lib/nimble_template/addons/github.ex
+++ b/lib/nimble_template/addons/github.ex
@@ -83,11 +83,6 @@ defmodule NimbleTemplate.Addons.Github do
   end
 
   @impl true
-  def do_apply(%Project{} = project, %{github_action_deploy_heroku: false}) do
-    project
-  end
-
-  @impl true
   def do_apply(%Project{} = project, %{github_wiki: true}) do
     project
     |> copy_wiki_files()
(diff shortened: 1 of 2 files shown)


lib/nimble_template/addons/github.ex
```diff
@@ -84,12 +83,12 @@ defmodule NimbleTemplate.Addons.Github do
   end
 
   @impl true
-  def do_apply(%Project{} = project, opts) when not opts.github_action_deploy_heroku do
+  def do_apply(%Project{mix_project?: true} = project, %{github_action_deploy_heroku: true}) do
```
rosle, line 86: I'm not sure if this is possible as on mix project we don't ask for Heroku deployment setup, right? 🤔

rosle (approved):

lib/nimble_template/addons/github.ex
andyduong1920: That is a good question Ros, let me check the Mix template, and the Github Addon a bit.

andyduong1920: adjusted in https://github.com/nimblehq/elixir-templates/pull/152/commits/0f9b3b719a6e7a64d3054956ad12959845b7cbb0 @rosle

bterone (approved):

Issue #105, opened 2021-05-18
Add CSS mixin


andyduong1920:
## Why

- text-ellipsis
- rem

## Who Benefits?

Developer

andyduong1920:
we can do that at the same time with https://github.com/nimblehq/elixir-templates/issues/100

Issue #100, opened 2021-05-10
Organize CSS structure to following Nimble's convention


andyduong1920:
## Why

To following Nimble's convention https://nimblehq.co/compass/development/code-conventions/css/#stylesheets-structure

## Who Benefits?

Nimbler

PR #154, opened 2022-01-05
Add NimbleCSS and NimbleJS addons

andyduong1920:
https://github.com/nimblehq/elixir-templates/issues/100
https://github.com/nimblehq/elixir-templates/issues/105

## What happened

Init`Nimble CSS` and `Nimble JS` structure.

### PostCSS

We need to set up the PostCSS as there was some error cause by the `stylelint-config-sass-guidelines`

### CSS

I bring 2 default files into the template, 1 is the `sizing` function and another one is the default layout

image

Feel free to let me know if that doesn't make sense or we need to adjust something.

### JS

I bring some files into the template, we use that mostly on our projects, feel free to let me know if that doesn't make sense or we need to adjust something.

image

## Proof Of Work

image


priv/templates/nimble_template/assets/js/helpers/storageAvailable.js
```diff
@@ -0,0 +1,46 @@
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API
+ */
+
+const STORAGE_TYPES = {
+  localStorage: "localStorage",
+};
+
+function isStorageAvailable(type) {
```
rosle, line 9: Not sure about this, are you seeing us using this a lot 🤔 ?

lib/nimble_template/addons/variants/phoenix/web/nimble_css.ex
```diff
@@ -0,0 +1,53 @@
...
+    project
+  end
+
+  defp edit_style_lint_rc(project) do
+    Generator.replace_content(
+      "assets/.stylelintrc.json",
+      """
+        "ignoreFiles": [
+          "css/app.css",
+          "css/phoenix.css"
```
rosle, line 43: Just curious what is the error and can it be fixed on post apply 💭  If it can be fixed, maybe we can just remove them from the base template 💭 💡

andyduong1920: There are many errors from these files as contains the error from Milligram's codebase  https://milligram.io/

And it can't fix by the `post apply` @rosle :D

priv/templates/nimble_template/assets/js/helpers/storageAvailable.js
andyduong1920: so we can remove it, right :D

priv/templates/nimble_template/assets/js/helpers/display.js
```diff
@@ -0,0 +1,10 @@
+/*
+ * Helper functions to show and hide elements
+ * */
+export function showElement(elementRef) {
+  elementRef.classList.remove("d--none");
```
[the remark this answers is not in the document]
andyduong1920: good catch Ros, I think better to remove it for now :D

lib/nimble_template/addons/variants/phoenix/web/core_js.ex
```diff
@@ -16,30 +16,17 @@ defmodule NimbleTemplate.Addons.Phoenix.Web.CoreJS do
...
-    end
+  def edit_package_json(%Project{} = project) do
+    Generator.replace_content(
+      "assets/package.json",
+      """
+        "dependencies": {
+      """,
+      """
+        "dependencies": {
+          "core-js": "^3.7.0",
```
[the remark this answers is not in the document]
andyduong1920: There is some downside with that solution Ros

1/ Harder to test this addon in the Unit test
2/ On applying, as we have different Node package add-ons (eg CoreJS, prettier, eslint,...) we will end up calling the `yarn add ...` multiple times -> connection to `npm registry multiples time` -> reduce the apply time.

So I will keep the current implementation for

1/ Easy for unit test
2/ We will call the `npm install or yarn install`  only once time at the `post apply`

priv/templates/nimble_template/assets/js/helpers/storageAvailable.js
andyduong1920: Removed in https://github.com/nimblehq/elixir-templates/pull/154/commits/ee7be53e5dbe91a94179262de72b85db22e61cb3

priv/templates/nimble_template/assets/js/helpers/display.js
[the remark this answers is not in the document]
andyduong1920: Removed in https://github.com/nimblehq/elixir-templates/pull/154/commits/ee7be53e5dbe91a94179262de72b85db22e61cb3

priv/templates/nimble_template/assets/css/mixins/_text-ellipsis.scss
```diff
@@ -0,0 +1,20 @@
+@mixin text-ellipsis(
+  $lines: 2,
+  $font-size: $font-size-base,
+  $line-height: $line-height-base
```
bterone, line 4: These variables come from bootstrap, shall we add them into the `_variables.scss` in case the user doesn't put bootstrap in the project? 🤔

rosle (approved):

priv/templates/nimble_template/assets/css/mixins/_text-ellipsis.scss
andyduong1920: I was thinking about that @bterone but I decide to not add that, because in case someone really would like to use that they will need to decide their `base font-size` and `base line-height` on their end.

Also this is the NimbleCSS addon, so the user is mostly coming from Nimble team :D and I'm assuming most of our project will use Boostrap so far :D 

But if you still think we need to define a `default value` so what value should we define as the base 🤔

bterone: For default values, `font-size` could be the same as what we defined in `_sizing.scss` `16px` 
and line height could be `1.5` from bootstrap's [variables](https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L558) for now, what do you think? 🤔 

Since this addon could also be used by the community, they might wonder why the mixin isn't working out of the box and they'd need to manually assign variable values 😱 

On a separate topic, there's also the [!default](https://sass-lang.com/documentation/variables#default-values) flag we could experiment with, perhaps it could be useful for this case 💭

topnimble (approved):

priv/templates/nimble_template/assets/css/mixins/_text-ellipsis.scss
andyduong1920: added in https://github.com/nimblehq/elixir-templates/pull/154/commits/0a69723890bc1ce079f220a54baf41673461f650 @bterone

bterone (approved):

PR #162, opened 2022-01-19
Refactor the Github action workflow 

andyduong1920:
## What happened

Refactor the Github action workflow to use the new Github Action reusable feature

## Insight

1/ https://github.blog/2021-11-29-github-actions-reusable-workflows-is-generally-available/
2/ The Reusable workflow must be in the `.github/workflows`
3/ Adjust the workflow name

### Before

image

### After 

image

### The test is pass

image


olivierobert (approved):

.github/workflows/apply_api_variant.yml
```diff
@@ -0,0 +1,32 @@
+name: Apply API variant
+
+on: push
+
+jobs:
+  standard_api_project:
+    name: Standard API project
+    uses: nimblehq/elixir-templates/.github/workflows/reusable_phoenix_project.yml@composite_2.0
```
longnd, line 8: I think we can simplify it like this, according to Github [document](https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow)
> `./{path}/{filename}` for reusable workflows in the same repository.

```suggestion
    uses: ./.github/workflows/reusable_phoenix_project.yml@composite_2.0
```

.github/workflows/test_template.yml
```diff
@@ -10,7 +10,7 @@ env:
 
 jobs:
   unit_test:
-    name: Test template
+    name: Unit test
```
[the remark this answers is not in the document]
andyduong1920: Updated in https://github.com/nimblehq/elixir-templates/pull/162/commits/0880c88156d60d69f92e16b236c346be4778a6d9 @longnd

.github/workflows/apply_api_variant.yml
```diff
@@ -0,0 +1,32 @@
+name: Apply API variant
+
+on: push
+
+jobs:
+  standard_api_project:
+    name: Standard API project
```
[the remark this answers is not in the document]
andyduong1920: Updated in https://github.com/nimblehq/elixir-templates/pull/162/commits/0880c88156d60d69f92e16b236c346be4778a6d9 @longnd

.github/workflows/apply_api_variant.yml
andyduong1920: I'm not sure about this but `./.github/workflows/reusable_phoenix_project.yml@composite_2.0` it doesn't work @longnd, the workflow doesn't start 😢

image

andyduong1920: image

andyduong1920: That why I have to revert it back to `nimblehq/elixir-templates/`

image

longnd: @andyduong1920 I think by referring to the file using a relative path, we can't specify the version, it means we can only do like this

```suggestion
    uses: ./.github/workflows/reusable_phoenix_project.yml@composite_2.0
```

but this isn't a big deal. I'm fine with the current approach 🤞

longnd: @andyduong1920 I think by referring to the file using a relative path, we can't specify the version, it means we can only do like this

```suggestion
    uses: ./.github/workflows/reusable_phoenix_project.yml@composite_2.0
```

but this isn't a big deal. I'm fine with the current approach 🤞

PR #163, opened 2022-01-20
Improve the Test workflow with Github action Cache

andyduong1920:
## What happened

Separate the Test workflow into different steps to make the Github Cache working effective 

## Insight

For the test workflow, it contains multiple steps

- install_and_compile_dependencies
- lint_codebase
- unit_test

We need to separate to different steps to have the benefit from the Github cache action, because the Github cache action doesn't cache if the `mix codebase is failed`, so on the next time, it won't have the cache.

image

With the install_and_compile_dependencies it just installs the dependency and caches it, so on the next step (and next time), it will use the cache.

### Github Addon

Refactor the Github Addon

1/ Mix project: Separate into 3 steps

2/ Phoenix project: Separate into 5 steps, also introduce the new `test seed data` on each PR, we have that in one of our client projects.

image

Allow the feature test using partitions mode.

## Proof Of Work

The test is passed.

image

### DIFF
--- a/.github/workflows/test_template.yml
+++ b/.github/workflows/test_template.yml
@@ -51,6 +51,49 @@ jobs:
 
       - name: Run codebase check
         run: mix codebase
+        
+  unit_test:
+    name: Unit test
+    
+    needs: lint_codebase
+    
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.8.0
+        with:
+          access_token: ${{ github.token }}
+
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+
+      - name: Set up Elixir ${{ env.ELIXIR_VERSION }} and OTP ${{ env.OTP_VERSION }}
+        uses: erlef/setup-elixir@v1
+        with:
+          otp-version: ${{ env.OTP_VERSION }}
+          elixir-version: ${{ env.ELIXIR_VERSION }}
+
+      - name: Cache Elixir build
+        uses: actions/cache@v2
+        with:
+          path: |
+            _build
+            deps
+          key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-mix-
+
+      - name: Install Dependencies
+        run: mix deps.get
+
+      - name: Compile dependencies
+        run: mix compile
+
+      - name: Install Phoenix ${{ env.PHOENIX_VERSION }}
+        run: make install_phoenix PHOENIX_VERSION=${{ env.PHOENIX_VERSION }}
 
       - name: Run Tests
         run: mix test
--- a/.github/workflows/test_template.yml
+++ b/.github/workflows/test_template.yml
@@ -9,11 +9,56 @@ env:
   MIX_ENV: test
 
 jobs:
-  unit_test:
-    name: Run unit test
+  install_and_compile_dependencies:
+    name: Install and Compile Dependencies
     runs-on: ubuntu-latest
 
     steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.8.0
+        with:
+          access_token: ${{ github.token }}
+
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+
+      - name: Set up Elixir ${{ env.ELIXIR_VERSION }} and OTP ${{ env.OTP_VERSION }}
+        uses: erlef/setup-elixir@v1
+        with:
+          otp-version: ${{ env.OTP_VERSION }}
+          elixir-version: ${{ env.ELIXIR_VERSION }}
+
+      - name: Cache Elixir build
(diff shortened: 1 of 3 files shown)


priv/templates/nimble_template/.github/workflows/test.yml.eex
```diff
@@ -87,22 +144,254 @@ jobs:
 
       - name: Run codebase check
         run: mix codebase
+    
+  seed_data:
```
olivierobert, line 148: What is this step for? For what type of tests is this required? Also, now since it does not appear as a dependency to any step, I am not sure what will is its use 🤔  Also isn't global test data an anti-pattern i.e., should we create only the required test data for each test?

.github/workflows/test_template.yml
```diff
@@ -43,6 +43,41 @@ jobs:
...
+        with:
+          path: |
+            _build
+            deps
+          key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-mix-
+
+      - name: Install Dependencies
+        run: mix deps.get
```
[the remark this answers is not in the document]
andyduong1920: We can't skip that step as I remember @olivierobert thus the cached is there and there is no new dependency added but we still need to trigger the `mix deps.get` so Mix can understand that `mix deps.get` was done, without that, it will raise the error like `dependency doesn't install` 

I will make a demo and then upload it here.

priv/templates/nimble_template/.github/workflows/test.yml.eex
andyduong1920: Hi @olivierobert this is not the test data, this is the `job for testing the seed script`, as the codebase/schema/migration is changed frequently and the developer does not run the `seed script` on development frequently, so with the codebase updated above, the seed script might be failed, that affect into the new people, who just join the project and running the seed, they need to spend time to fix the seed script.

So with that reason, we decided to have the `test for the seed script` - this job is separated with the `unit test`, here we only test the `seed script` to make sure the seed script is able to run, so the new joiner will happy with that.

The cost for this test is really small, just 10 seconds in most of standard projects, that doesn't take too much CI time, but it bring the benefit to the new joiner.

.github/workflows/test_template.yml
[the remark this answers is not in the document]
andyduong1920: I have added a commit to test it.

1/ Commit

image

2/ Result 

Install dependency is run on the first step - Install and Compile Dependencies 

image

On the second step (without the `ix deps.get` step above) it failed

image

priv/templates/nimble_template/.github/workflows/test.yml.eex
andyduong1920: This is an example where the developer modified the schema field name but didn't update the seed script @olivierobert.

image

The result is this seed script won't run :D and the new joiner or who needs to re-run the `seed script` will fix that 😂.

But with the `test seed_script` here, we can avoid that case as the CI will be failed and the modifier need to adjust the seed script along with the ecto schema field too ✅

olivierobert: Well understood. Indeed, the seeds are not usually executed that often so ensuring these still run is a good idea. In terms of CI time, it is reasonable. The only downside I see is the amount of workflow code needed to test something small 🙈 

I will add a few suggestions to avoid some confusion like I had.

priv/templates/nimble_template/.github/workflows/test.yml.eex
```diff
@@ -87,22 +144,254 @@ jobs:
...
+    name: Test seed data
```
olivierobert, line 149: To make it clearer, I recommend naming this:

```suggestion
    name: Test database seeds
```

It is a small change that makes it clear IMO

priv/templates/nimble_template/.github/workflows/test.yml.eex
```diff
@@ -87,22 +144,254 @@ jobs:
...
+      - name: Compile assets
+        run: npm run --prefix assets build:dev
+      <% end %>
+      - name: Create database
+        run: mix ecto.create
+
+      - name: Migrate database
+        run: mix ecto.migrate
+
+      - name: Seed data
```
olivierobert, line 224: How about:

```suggestion
      - name: Seed database
```

.github/workflows/test_template.yml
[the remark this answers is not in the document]
olivierobert: All clear 👍

priv/templates/nimble_template/.github/workflows/test.yml.eex
andyduong1920: Updated in https://github.com/nimblehq/elixir-templates/pull/163/commits/8c6197c1f715061cc3261c1ac2c284a90c71e079

priv/templates/nimble_template/.github/workflows/test.yml.eex
andyduong1920: Updated in https://github.com/nimblehq/elixir-templates/pull/163/commits/8c6197c1f715061cc3261c1ac2c284a90c71e079

longnd (approved):

topnimble (approved):

PR #164, opened 2022-01-20
Finalize the Base API setup feature 

andyduong1920:
https://github.com/nimblehq/elixir-templates/issues/127

## What happened

Finalize the Base API setup feature 

## Insight

- Add alias MODULE.ParamsValidator into web entry point.

## Proof Of Work

The test is passed

### DIFF
--- a/lib/nimble_template/addons/variants/phoenix/api/params_validation.ex
+++ b/lib/nimble_template/addons/variants/phoenix/api/params_validation.ex
@@ -3,7 +3,43 @@ defmodule NimbleTemplate.Addons.Phoenix.Api.ParamsValidation do
 
   use NimbleTemplate.Addons.Addon
 
-  def do_apply(%Project{} = project, _opts), do: copy_files(project)
+  def do_apply(%Project{} = project, _opts) do
+    project
+    |> copy_files()
+    |> edit_files()
+  end
+
+  defp edit_files(project) do
+    edit_web_entrypoint(project)
+  end
+
+  defp edit_web_entrypoint(%Project{web_module: web_module, web_path: web_entrypoint} = project) do
+    Generator.replace_content(
+      "#{web_entrypoint}.ex",
+      """
+        def controller do
+          quote do
+            use Phoenix.Controller, namespace: #{web_module}
+
+            import Plug.Conn
+            import #{web_module}.Gettext
+            alias #{web_module}.Router.Helpers, as: Routes
+      """,
+      """
+        def controller do
+          quote do
+            use Phoenix.Controller, namespace: #{web_module}
+
+            import Plug.Conn
+            import #{web_module}.Gettext
+
+            alias #{web_module}.ParamsValidator
+            alias #{web_module}.Router.Helpers, as: Routes
+      """
+    )
+
+    project
+  end
 
   defp copy_files(
          %Project{
--- a/test/nimble_template/addons/variants/phoenix/api/params_validation_test.exs
+++ b/test/nimble_template/addons/variants/phoenix/api/params_validation_test.exs
@@ -14,5 +14,18 @@ defmodule NimbleTemplate.Addons.Phoenix.Api.ParamsValidationTest do
         assert_file("test/nimble_template_web/params/params_validator_test.exs")
       end)
     end
+
+    test "adds ParamsValidator alias into the web entrypoint", %{
+      project: project,
+      test_project_path: project_path
+    } do
+      in_test_project(project_path, fn ->
+        ApiAddons.ParamsValidation.apply(project)
+
+        assert_file("lib/nimble_template_web.ex", fn file ->
+          assert file =~ "alias NimbleTemplateWeb.ParamsValidator"
+        end)
+      end)
+    end
   end
 end


test/nimble_template/addons/variants/phoenix/api/json_api_test.exs
```diff
@@ -0,0 +1,33 @@
...
+      in_test_project(test_project_path, fn ->
+        ApiAddons.JsonApi.apply(project)
+
+        assert_file("mix.exs", fn file ->
+          assert file =~ "{:jsonapi, \"~> 1.3.0\"}"
+        end)
+      end)
+    end
+
+    test "adds config for json_api in config/config.exs", %{
```
byhbt, line 20:
```suggestion
    test "adds config for jsonapi in config/config.exs", %{
```

byhbt (approved):

test/nimble_template/addons/variants/phoenix/api/json_api_test.exs
andyduong1920: Thanks @byhbt I have changed it in https://github.com/nimblehq/elixir-templates/pull/164/commits/795704734b550b36320749f0321e8ee64f0d8fc7

lib/nimble_template/addons/variants/phoenix/api/empty_body_plug.ex
```diff
@@ -0,0 +1,54 @@
+defmodule NimbleTemplate.Addons.Phoenix.Api.EmptyBodyPlug do
```
junan, line 1: Can we rename it the same as the file name? `EmptyBodyParamsPlug`

lib/nimble_template/addons/variants/phoenix/api/params_validation.ex
```diff
@@ -3,7 +3,43 @@ defmodule NimbleTemplate.Addons.Phoenix.Api.ParamsValidation do
...
+    project
+    |> copy_files()
+    |> edit_files()
+  end
+
+  defp edit_files(project) do
+    edit_web_entrypoint(project)
+  end
+
+  defp edit_web_entrypoint(%Project{web_module: web_module, web_path: web_entrypoint} = project) do
```
junan, line 16: O think the corrct name is [`entry point`](https://en.wikipedia.org/wiki/Entry_point)
```suggestion
  defp edit_web_entry_point(%Project{web_module: web_module, web_path: web_entrypoint} = project) do
```

priv/templates/nimble_template/lib/otp_app_web/controllers/api/fallback_controller.ex.eex
```diff
@@ -0,0 +1,13 @@
+defmodule <%= web_module %>.Api.FallbackController do
+  use Phoenix.Controller
+
+  alias Ecto.Changeset
+  alias <%= web_module %>.Api.ErrorView
+
+  def call(conn, {:error, :invalid_params, %Changeset{valid?: false} = changeset}) do
```
junan, line 7: Just a question, when this will be called?

priv/templates/nimble_template/lib/otp_app_web/plugs/check_empty_body_params_plug.ex.eex
```diff
@@ -0,0 +1,23 @@
...
+  import Phoenix.Controller, only: [put_view: 2, render: 3]
+  import Plug.Conn
+
+  alias <%= web_module %>.Api.ErrorView
+
+  @apply_for_methods ["POST", "PUT", "PATCH"]
+
+  def init(opts), do: opts
+
+  def call(%{body_params: body_params, method: method} = conn, _opts)
```
junan, line 13: Are we checking here that `post/put/patch` request body empty or not, if it is empty will render an error, is it?
 
If it then I am just wondering if the patch API is like `patch https://example.com/event/:id/cancel` where it will be just to cancel the event and in this case, there might not be needed no any request body params due to URL. So in this case our app will return an error.

I think the main question is about our API conventions, 
-  Should we make `request body` mandatory for every `post/put/patch` API endpoint
-  Should we make `request body` optional for every `post/put/patch and others` API endpoint

priv/templates/nimble_template/lib/otp_app_web/controllers/api/fallback_controller.ex.eex
andyduong1920: When the params is invalid on the `controller params` level @junan you can check it in the `params_validator.ex`

priv/templates/nimble_template/lib/otp_app_web/plugs/check_empty_body_params_plug.ex.eex
andyduong1920: Good point @junan I think for now we only validate the POST request only, the PATCH could have an empty body. What do you think @byhbt

lib/nimble_template/addons/variants/phoenix/api/empty_body_plug.ex
andyduong1920: good catch, updated in https://github.com/nimblehq/elixir-templates/pull/164/commits/43577587fd1e055db9901b8bd2ec2d1f119c2fb3

priv/templates/nimble_template/lib/otp_app_web/plugs/check_empty_body_params_plug.ex.eex
byhbt: yup, good catch Junan, I agree with the solution we validate for the POST request only.

andyduong1920: I have updated it in https://github.com/nimblehq/elixir-templates/pull/164/commits/6a79d4d029614e6ed5d5e5c2b067b68f391898ad FYi @junan @byhbt, so we are checking only for the POST request

junan (approved):

Issue #120, opened 2021-07-29
Support Phoenix 1.6


andyduong1920:
## Why

As Phoenix 1.6 replaces all node, npm, webpack by esbuild

https://github.com/phoenixframework/phoenix/pull/4377

PR #165, opened 2022-02-14
Support Phoenix 1.6.x

andyduong1920:
https://github.com/nimblehq/elixir-templates/issues/120

## What happened

Support Phoenix 1.6.x.

## Insight

1/ Change the build script:
- On Phoenix 1.6, the default `mix phx.new` will generate a LiveView project.
- To generate a non-LiveView project, we need to pass the `--no-live`.
- Replace `--no-webpack` with `--no-assets` for the API project.

2/ Support the latest Phoenix 1.6.6

3/ Edit the addons, some highlights:
- The `config/runtime.ex` file is generated by default on Phoenix 1.6.x
- CoreJS -> remove the `import "regenerator-runtime/runtime"` as we don't use Babel anymore.
- PostCSS -> removed as we don't use Webpack anymore
- Using esbuild for JS precompiled.
- Using Dart-SASS for CSS precompiled (https://blog.mnishiguchi.com/elixir-phoenix-16-esbuild-scss)

image

- Generate the `package.json` file to contain the Node dependencies like
  - CoreJS
  - Sprite SVG
  - Stylelint
 - Update the Dockerfile.
  - Fixing the dart-sass issue on Alpine (https://github.com/CargoSense/dart_sass#compatibility-with-alpine-linux-mix-sass-default-exited-with-2)
  - Running `mix asserts.deploy` task 

## Proof Of Work

All test are passed


.github/workflows/test_template.yml
```diff
@@ -5,7 +5,7 @@ on: push
 env:
   OTP_VERSION: 24.0.4
   ELIXIR_VERSION: 1.12.2
```
longnd, line 7: since we're upgrading phoenix, should we upgrade Elixir & OTP to the latest version?
```suggestion
  OTP_VERSION: 24.2.1
  ELIXIR_VERSION: 1.13.3
```

Makefile
```diff
@@ -21,9 +21,10 @@ create_mix_project:
 common_addon_prompts = Y\nY\nY\nY\nY\nY\nY\nY\n
 
 # Y - in response to Would you like to add the SVG Sprite addon?
-# Y - in response to Would you like to add the Nimble CSS addon?
+# Y - in response to Would you like to add the Dart Sass addon?
```
longnd, line 24: I think in any web project, we will always need SCSS right so Dart SCSS is a must to compile it and should be a default addon. No?

lib/nimble_template/addons/variants/phoenix/web/core_js.ex
```diff
@@ -24,7 +24,7 @@ defmodule NimbleTemplate.Addons.Phoenix.Web.CoreJS do
       """,
       """
         "dependencies": {
-          "core-js": "^3.7.0",
+          "core-js": "^3.7.0"
```
longnd, line 27: please consider using the latest version https://www.npmjs.com/package/core-js/v/3.21.1
```suggestion
          "core-js": "^3.21.1"
```

.github/workflows/test_template.yml
andyduong1920: I think we should do that in a separate PR @longnd as this PR is only supporting Phoenix 1.6.x, and also Phoenix 1.6.x work with Erlang 24.0.4 and Elixir 1.12.2, we no need to introduce that in a same PR.

Makefile
andyduong1920: It depends @longnd for example the Micky and Thanh's project is using Tailwind CSS which doesn't require SCSS

lib/nimble_template/addons/variants/phoenix/web/core_js.ex
andyduong1920: It's not a part of this PR I think @longnd, we can create an issue to upgrade all Node dependency version after merge this PR

.github/workflows/test_template.yml
andyduong1920: we can create an issue to upgrade Erlang and Elixir version after merging this PR, what do you think @longnd

topnimble (approved):

.github/workflows/test_template.yml
andyduong1920: I have created another PR for that https://github.com/nimblehq/elixir-templates/pull/169 on the new Elixir version, they released a new ExUnit coverage algorithm so the test coverage is reduced, we need to spend time to fix that FYI @longnd, but with the Elixir 1.12.2, everything is good to go.

lib/nimble_template/addons/variants/phoenix/web/core_js.ex
andyduong1920: same issue on the upgrade Elixir and Erlang version, update each Node dependency might have some affect (the new version doesn't work) so prefer to do that on a separate PR @longnd

Makefile
longnd: it makes sense to me now 👍

longnd (approved):

andyduong1920:
Thanks @longnd and @topnimble 🔥

PR #166, opened 2022-02-17
Change the Github reusable to relative path.

andyduong1920:
https://github.com/nimblehq/elixir-templates/pull/162#discussion_r804546686

## What happened

Change the reusable workflow to a relative path followed by the suggestion from Long on https://github.com/nimblehq/elixir-templates/pull/162#discussion_r804546686

## Insight

https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow

With the relative path, we no need to release any tag for the reusable workflow like before, for now, it will always refer to the relative path on the same branch 😍 much much easier to understand and control when we need to adjust the `reusable workflow`.

## Proof Of Work

The test is passed.

### DIFF
--- a/.github/workflows/apply_api_variant.yml
+++ b/.github/workflows/apply_api_variant.yml
@@ -5,28 +5,28 @@ on: push
 jobs:
   standard_api_project:
     name: Test on a Standard API project
-    uses: nimblehq/elixir-templates/.github/workflows/reusable_phoenix_project.yml@composite_2.0
+    uses: ./.github/workflows/reusable_phoenix_project.yml
     with:
       new_project_options: "--no-html --no-webpack"
       variant: "api"
 
   non-standard_api_project:
     name: Test on a Non-Standard API project
-    uses: nimblehq/elixir-templates/.github/workflows/reusable_phoenix_project.yml@composite_2.0
+    uses: ./.github/workflows/reusable_phoenix_project.yml
     with:
       new_project_options: "--no-html --no-webpack --module=SampleCustomModule --app=sample_custom_app"
       variant: "api"
 
   standard_web_project:
     name: Test on a Standard Web project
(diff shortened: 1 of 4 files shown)


olivierobert (approved):
Much easier to use indeed.

It would be good to clean up/remove the unused tags afterward:

Screen Shot 2565-02-17 at 14 21 34

andyduong1920:
> Much easier to use indeed.
> 
> It would be good to clean up/remove the unused tags afterward:
> 
> Screen Shot 2565-02-17 at 14 21 34

Sure, this comes from the new version of `Git reusable workflow`, in the original version where it called `Github composite` it doesn't accept the relative path.

Really happy to see this new feature, it means now the reusable is really to use, we can think about it in some of our projects now.

Actually, we can have some base setup like Setup Elixir, Node, ... in 1 file and reuse it across many different job with the relative path, I will make another PR for that.

And sure, will clean the old `composite_x` releases once we have this PR merged to main :)

longnd (approved):

andyduong1920:
> > Much easier to use indeed.
> > It would be good to clean up/remove the unused tags afterward:
> > Screen Shot 2565-02-17 at 14 21 34
> 
> Sure, this comes from the new version of `Git reusable workflow`, in the original version where it called `Github composite` it doesn't accept the relative path.
> 
> Really happy to see this new feature, it means now the reusable is really to use, we can think about it in some of our projects now.
> 
> Actually, we can have some base setup like Setup Elixir, Node, ... in 1 file and reuse it across many different job with the relative path, I will make another PR for that.
> 
> And sure, will clean the old `composite_x` releases once we have this PR merged to main :)

@olivier FYI, I have tried that on https://github.com/nimblehq/elixir-templates/pull/167/files, we can compile `uses` and `steps` 

image

I was trying to have something like this

image

Hopefully the next version Github can let us define like that, then we can save many lines in the setup dependency across many jobs.

Issue #184, opened 2022-03-28
Fix: Replace config error when applying template


bterone:
## Issue

Raised a File Error when trying to apply the API template

![Screen Shot 2022-03-28 at 4 39 07 PM](https://user-images.githubusercontent.com/34730459/160374804-3ca612c2-6e80-42a1-9f04-345199f28ca0.png)

It seems some behavior is incorrectly modifying the `config/develop.exs` instead of `config/dev.exs`. Will require further investigation.
 
## Expected

Correctly apply replacement behavior to `config/dev.exs`.
 
## Steps to reproduce

1. Create a new phoenix project with `mix phx.new`
2. Add `{:nimble_template, "~> 4.1", only: :dev, runtime: false},` to `deps` in `mix.exs` as per instructions
3. `mix do deps.get, deps.compile`
4. Apply API variant `mix nimble_template.gen --api`, no optional addons are necessary.
5. Observe the error during the apply template step.

andyduong1920:
I also saw that, it's generate from the 3rd library, hopefully in the upcoming version they will fix that :D

bterone:
Oh, do we know from which 3rd party lib? @andyduong1920 👀

andyduong1920:
Actually I found this error, changed it in https://github.com/nimblehq/elixir-templates/pull/188

Issue #182, opened 2022-03-28
Fix: Formatting issues when creating new API project


bterone:
## Issue

When creating a new project with the latest Phoenix Installer, the project runs into an issue with `prod.exs` when running `mix format`.

![Screen Shot 2022-03-28 at 12 44 14 PM](https://user-images.githubusercontent.com/34730459/160333751-1af0f842-04ff-4d61-8843-5a3721e5ab27.png)

Versions:
- Erlang/OTP 24 [erts-12.2.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]
- Elixir 1.13.3 (compiled with Erlang/OTP 24)
- Phoenix Installer [`mix phx.new` ver.] v1.6.6
 
## Expected

The project can be formatted with no issues.
 
## Steps to reproduce

1. Create a new phoenix project with `mix phx.new` with the latest phoenix installer (1.6.6)
2. Add `{:nimble_template, "~> 4.1", only: :dev, runtime: false},` to `deps` in `mix.exs` as per instructions
3. `mix do deps.get, deps.compile`
4. Apply API variant `mix nimble_template.gen --api`, no addons are necessary.
5. Run `mix format`
6. Syntax error with `config/prod.exs` file.

andyduong1920:
not sure as I can't reproduce it 🤔

![image](https://user-images.githubusercontent.com/11751745/160386866-f7f97c2e-c2ae-4f67-a908-a980c2162044.png)


Another thing, on the test the Apply the variant workflows we have the `Run mix codebase` step for the new generated application, we don't have any error on the `mix format` step there @bterone 

![image](https://user-images.githubusercontent.com/11751745/160387081-f146775d-0122-41e7-8209-e436b24f538a.png)

![image](https://user-images.githubusercontent.com/11751745/160387452-b6e940d8-a983-436b-a317-1b57b920e68e.png)

https://github.com/nimblehq/elixir-templates/runs/5718992570?check_suite_focus=true

bterone:
What is the version of your `mix phx.new`? 🤔  @andyduong1920, is it 1.6.6?
I can always reproduce this error.

Also the `mix codebase` is our alias, it runs `credo, deps.unlock --check-unused` and `format`

andyduong1920:
> What is the version of your `mix phx.new`? 🤔 @andyduong1920, is it 1.6.6? I can always reproduce this error.
> 
[...]

PR #188, opened 2022-03-28
Fix: Formatting issues when creating new API project

andyduong1920:
Close https://github.com/nimblehq/elixir-templates/issues/182
Close https://github.com/nimblehq/elixir-templates/issues/184

## What happened

Fix the invalid format in `config/prod.exs`.

Also close the error on https://github.com/nimblehq/elixir-templates/issues/182

## Insight

When the project name is NOT long enough, the `config :#{otp_app}, #{web_module}.Endpoint,` is on 1 line, not on 2 lines, so for that, we can't use the triple quotes.

"""
config :#{otp_app}, #{web_module}.Endpoint,
"""

Change to use `"config :#{otp_app}, #{web_module}.Endpoint,"`, treat as normal string

## Proof Of Work

The test is passed

Tested on my local.


lib/nimble_template/addons/variants/phoenix/api/config.ex
```diff
@@ -17,12 +17,8 @@ defmodule NimbleTemplate.Addons.Phoenix.Api.Config do
...
     Generator.replace_content(
       "config/prod.exs",
-      """
-      config :#{otp_app}, #{web_module}.Endpoint,
-      """,
-      """
-      config :#{otp_app}, #{web_module}.Endpoint.Anchor
-      """
+      "config :#{otp_app}, #{web_module}.Endpoint,",
+      "config :#{otp_app}, #{web_module}.Endpoint.Anchor"
```
bterone, line 21: Hi @andyduong1920, I'm just a bit curious why we're using this approach to replace the content with `.Anchor` only to delete it in line 24. Is there a reason? 💭

andyduong1920: I have refactored it in https://github.com/nimblehq/elixir-templates/pull/188/commits/5e1006717fcffa9b6e4b0c4acce7049c9b8092f3 @bterone, please check the new version :)

bterone (approved):

longnd (approved):

PR #228, opened 2022-06-09
Add the VersionUpgradeHelper

andyduong1920:
## What happened

As we upgrade the version frequently, the process is quite boring, I have created a helper for that, so we just need to run that method and everything will be done.

The next step will be:

- Create the Github Action which receives up to 4 inputs (elixir, erlang, node, alpine version)
- That Github Action will upgrade the version similar to the [Bump Version](https://github.com/nimblehq/elixir-templates/blob/develop/.github/workflows/bump_version.yml)
- That Github Action will create a PR to upgrade those versions automatically, similar to the [Bump Version PR](https://github.com/nimblehq/elixir-templates/pull/227)
- Also the next step we will create a cronjob task on Github Action to fetch the latest Elixir, Erlang. Node and Alpine versions, run this task and create a PR automatically 🤞

## Insight

Add the `upgrade_elixir_erlang_node_and_alpine` in the Version module

## Proof Of Work

The test is passed

The command is used to generate the https://github.com/nimblehq/elixir-templates/pull/229

image

### DIFF
--- a/lib/nimble_template/helpers/generator.ex
+++ b/lib/nimble_template/helpers/generator.ex
@@ -24,7 +24,7 @@ defmodule NimbleTemplate.Generator do
 
   def rename_file(old_path, new_path), do: File.rename(old_path, new_path)
 
-  def replace_content(file_path, anchor, content) do
+  def replace_content(file_path, anchor, content, raise_exception \\ true) do
     file = Path.join([file_path])
 
     file_content =
@@ -40,7 +40,21 @@ defmodule NimbleTemplate.Generator do
         File.write!(file, [left, content, right])
 
       :error ->
-        Mix.raise(~s[Could not find #{anchor} in #{file_path}])
+        if raise_exception do
+          Mix.raise(~s[Could not find #{anchor} in #{file_path}])
+        else
+          :error
+        end
+    end
+  end
+
+  def replace_content_all(file_path, anchor, content) do
+    case replace_content(file_path, anchor, content, false) do
+      :ok ->
+        replace_content_all(file_path, anchor, content)
+
+      :error ->
+        nil
     end
   end
 
--- a/lib/nimble_template/version.ex
+++ b/lib/nimble_template/version.ex
@@ -64,6 +64,12 @@ defmodule NimbleTemplate.Version do
       "@elixir_version \"#{new_version}\""
     )
 
+    Generator.replace_content(
+      ".tool-versions",
+      "elixir #{current_version}",
+      "elixir #{new_version}"
+    )
+
     Generator.replace_content_all(
       "test/nimble_template/addons/asdf_tool_version_test.exs",
       "elixir #{current_version}-otp-#{}",
@@ -96,6 +102,18 @@ defmodule NimbleTemplate.Version do
       "lib/nimble_template/projects/project.ex",
       "@erlang_version \"#{current_version}\"",
       "@erlang_version \"#{new_version}\""
+    )
+
+    Generator.replace_content(
+      ".tool-versions",
+      "erlang #{current_version}",
+      "erlang #{new_version}"
+    )
+
+    Generator.replace_content(
+      ".tool-versions",
+      "-otp-#{Project.get_otp_major_version(current_version)}",
+      "-otp-#{Project.get_otp_major_version(new_version)}"
     )
 
     Generator.replace_content_all(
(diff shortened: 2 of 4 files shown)


lib/nimble_template/helpers/generator.ex
```diff
@@ -24,7 +24,7 @@ defmodule NimbleTemplate.Generator do
 
   def rename_file(old_path, new_path), do: File.rename(old_path, new_path)
 
-  def replace_content(file_path, anchor, content) do
+  def replace_content(file_path, anchor, content, raise_exception \\ true) do
```
olivierobert, line 27: Shouldn't we instead define another bang function? It is the same for [Enum.fetch/2](https://hexdocs.pm/elixir/Enum.html#fetch/2) and [Enum.fetch!/2](https://hexdocs.pm/elixir/Enum.html#fetch!/2)

```suggestion
  def replace_content(file_path, anchor, content) # existing function
  
  def replace_content!(file_path, anchor, content) # raises an exception
```

This So post is [a good recap](https://stackoverflow.com/questions/32449423/whats-the-meaning-of-and-syntax-in-elixir/32456168#32456168). It is the same convention in Ruby 🤗

lib/nimble_template/version.ex
```diff
@@ -15,6 +16,31 @@ defmodule NimbleTemplate.Version do
     end
   end
 
+  def upgrade_elixir_erlang_node_and_alpine(versions) do
```
olivierobert, line 19: This function name is very descriptive but quite long. How about having a shorter name but moving the descriptive part to the argument:

```suggestion
  def upgrade_stack(elixir_erlang_node_alpine_versions) do
```

lib/nimble_template/helpers/generator.ex
andyduong1920: True, that makes sense, I forgot that convention will make that change :)

andyduong1920: Actually, we would need to change for many places because the `existing function` raises the exception :D Will open it on a separate PR :D

lib/nimble_template/version.ex
andyduong1920: Renamed the method in https://github.com/nimblehq/elixir-templates/pull/228/commits/0065a70cc1a28b8c3e0a440c6cb12cc4c284586e

lib/nimble_template/helpers/generator.ex
andyduong1920: I will create a separate issue for this FYI @olivierobert

andyduong1920: Issue created https://github.com/nimblehq/elixir-templates/issues/230

lib/nimble_template/helpers/generator.ex
```diff
@@ -40,7 +40,21 @@ defmodule NimbleTemplate.Generator do
...
+        if raise_exception do
+          Mix.raise(~s[Could not find #{anchor} in #{file_path}])
+        else
+          :error
+        end
+    end
+  end
+
+  def replace_content_all(file_path, anchor, content) do
+    case replace_content(file_path, anchor, content, false) do
```
Nihisil, line 52: Minor comment. Looking at this line it is unclear what exactly `false` arg is intended for. We would need to check the function definition to find out that. Maybe tweak it like this? Since there are no named arguments in Elixir (well there are, but they requires to use maps).
```suggestion
    raise_exception = false
    case replace_content(file_path, anchor, content, raise_exception) do
```

andyduong1920: Hi @Nihisil with the @olivierobert suggestion above, we will have separate method

1. replace_content -> which won't raise any exception
2. replace_content! -> which will raise the exception

So we will remove the `false` argument here, I planed to refactor it on the different PR, is that ok?

Nihisil: Sure, it is totally fine. Thanks!

Nihisil (approved):

lib/nimble_template/helpers/generator.ex
andyduong1920: Noted ✅

olivierobert (approved):

lib/nimble_template/helpers/generator.ex
andyduong1920: Changed the name in https://github.com/nimblehq/elixir-templates/pull/288 👌

Issue #207, opened 2022-05-19
Add formatter configuration for .heex files


Nihisil:
## Why

It is convenient to have a default formatter for `.heex` files as well, so you don't need to do it manually (or use a third-party tool).

Formatting requires Elixir `1.13.4`, which is greater than this template supported `1.13.3` version.

It was added in `phoenix_live_view`, `0.17.8` version: https://github.com/phoenixframework/phoenix_live_view/blob/master/CHANGELOG.md#0178-2022-04-06

Instructions on how to set up you can find here: https://github.com/phoenixframework/phoenix_live_view/blob/master/lib/phoenix_live_view/html_formatter.ex

So, we can add this formatter feature once we update the template to the new Elixir version.

## Who Benefits?

Developers! Developers! Developers!

andyduong1920:
@Nihisil FYI the https://github.com/nimblehq/elixir-templates/pull/229 will upgrade Elixir to `1.13.4` 

but to solve this issue, we need some tweaks in the `.formatter.exs` file to make that work

![image](https://user-images.githubusercontent.com/11751745/172831010-5f71a3bd-d0f0-445f-a3b2-da0186eee2c7.png)

Will keep you posted

PR #229, opened 2022-06-09
Upgrade Elixir, Erlang and Alpine version

andyduong1920:
## What happened

Upgrade 

- Elixir
- Erlang
- Alpine

Add the HeexFormatter -> Closes https://github.com/nimblehq/elixir-templates/issues/207

## Insight

This is done by the helper on https://github.com/nimblehq/elixir-templates/pull/228

image

Note:

I also update the `test/nimble_template/version_test.exs` to use a different version format from the actual version as it could be a conflict between the real version vs the test version.

## Proof Of Work

The test is passed

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
-erlang 24.2.2
-elixir 1.13.3-otp-24
+erlang 25.0.1
+elixir 1.13.4-otp-25
 nodejs 16.15.0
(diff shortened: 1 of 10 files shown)


nvminhtue (approved):

Nihisil (approved):
Thanks for adding .heex files formatter!

Issue #198, opened 2022-04-27
Error during the generation of files on Phoenix version 1.6.7


liamstevens111:
## Issue

Error raised during the command 

```mix nimble_template.gen --web```

with the resulting error:

``` (Mix) Can't read test/support/channel_case.ex```

<img width="1291" alt="Screenshot 2565-04-27 at 12 54 44" src="https://user-images.githubusercontent.com/8955671/165466808-2cbb8bad-16a2-45ce-ac06-f145df7ec0bb.png">

The file **test/support/channel_case.ex** seems to have been removed in Phoenix version 1.6.7 and so the generation of files and subsequent questions do not occur.

[Link showing the Diff/Removal](https://www.phoenixdiff.org/?source=1.6.6&target=1.6.7)

## Expected
 
Questions are asked and files are generated according to the Nimble web template

## Steps to reproduce

1.  Generate a new Phoenix project with version 1.6.7 using **mix phx.new**
2. Add **{:nimble_template, "~> 4.1", only: :dev, runtime: false}** to mix.exs deps
3. Run **mix do deps.get, deps.compile**
4. Run **mix nimble_template.gen --web**

andyduong1920:
Thanks for raising this @liamstevens111 it's because currently the template is tested on Phoenix 1.6.6 :) 

<img width="698" alt="image" src="https://user-images.githubusercontent.com/11751745/165469809-750d7319-ee4e-4da3-9786-f0de4deaadb8.png">

byhbt:
Should we upgrade it to support the latest version of Phoenix `1.6.11` instead of just supporting the version `1.6.7` @andyduong1920 ?

andyduong1920:
> Should we upgrade it to support the latest version of Phoenix `1.6.11` instead of just supporting the version `1.6.7` @andyduong1920 ?

Sure, that make sense @byhbt 👍

PR #250, opened 2022-08-28
[gh-198] Fix error during the generation of files on Phoenix version 1.6.7

byhbt:
https://github.com/nimblehq/elixir-templates/issues/198

## What happened 👀

- [x] Update the test Phoenix version in the Github workflow
- [x] Update the README content for supporting the Phoenix `1.6.11`
- [x] Remove the modification related to the `channel_case.ex`

## Insight 📝

Remove the scripts updating the `channel_case.ex`. Because of the changes in code structure when upgrading from Phoenix `1.6.6` -> `1.6.7` and later versions.

Reference diff: https://www.phoenixdiff.org/?source=1.6.6&target=1.6.11

## Proof Of Work 📹

The test should pass when using the latest Phoenix version `1.6.11`

### DIFF
--- a/lib/nimble_template/addons/test_env.ex
+++ b/lib/nimble_template/addons/test_env.ex
@@ -133,10 +133,7 @@ defmodule NimbleTemplate.Addons.TestEnv do
   end
 
   defp edit_test_support_cases(project) do
-    project
-    |> edit_test_support_case("channel_case")
-    |> edit_test_support_case("conn_case")
-    |> edit_test_support_case("data_case")
+    edit_test_support_case(project, "data_case")
   end
 
   defp edit_test_support_case(project, support_case_name) do
--- a/lib/nimble_template/addons/variants/phoenix/mix_release.ex
+++ b/lib/nimble_template/addons/variants/phoenix/mix_release.ex
@@ -27,33 +27,6 @@ defmodule NimbleTemplate.Addons.Phoenix.MixRelease do
   end
 
   defp edit_files(%{otp_app: otp_app, web_module: web_module} = project) do
-    Generator.delete_content(
-      "config/runtime.exs",
-      """
-      # Start the phoenix server if environment is set and running in a release
-      if System.get_env("PHX_SERVER") && System.get_env("RELEASE_NAME") do
-        config :#{otp_app}, #{web_module}.Endpoint, server: true
-      end
-
-      """
-    )
-
-    Generator.delete_content(
-      "config/runtime.exs",
-      """
-
-        # ## Using releases
-        #
-        # If you are doing OTP releases, you need to instruct Phoenix
-        # to start each relevant endpoint:
-        #
-        #     config :#{otp_app}, #{web_module}.Endpoint, server: true
-        #
-        # Then you can assemble a release by calling `mix release`.
-        # See `mix help release` for more information.
-      """
-    )
-
     Generator.replace_content(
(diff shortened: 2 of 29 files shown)


lib/nimble_template/addons/test_env.ex
```diff
@@ -133,10 +133,7 @@ defmodule NimbleTemplate.Addons.TestEnv do
   end
 
   defp edit_test_support_cases(project) do
-    project
-    |> edit_test_support_case("channel_case")
-    |> edit_test_support_case("conn_case")
-    |> edit_test_support_case("data_case")
+    edit_test_support_case(project, "data_case")
```
andyduong1920, line 136: I guess we still have the `conn_case` on the Phoenix 1.6.11 🤔should we keep the `edit_test_support_cases(conn_case)` there @byhbt

lib/nimble_template/addons/variants/phoenix/mix_release.ex
```diff
@@ -27,33 +27,6 @@ defmodule NimbleTemplate.Addons.Phoenix.MixRelease do
   end
 
   defp edit_files(%{otp_app: otp_app, web_module: web_module} = project) do
-    Generator.delete_content(
-      "config/runtime.exs",
-      """
-      # Start the phoenix server if environment is set and running in a release
```
andyduong1920, line 33: should we keep deleting these content on Phoenix 1.6.11 @byhbt the purpose of this is delete the default config from Phoenix after generate owr project 🤔

byhbt: Yup, good catch, i revert the delete content here: https://github.com/nimblehq/elixir-templates/pull/250/commits/ea1c5ea6fc4df5c9561fa9badba7bb4337f3881b

However on the new version, the config content is changed. You can check here: https://www.phoenixdiff.org/?source=1.6.6&target=1.6.11 at `config/runtime.exs`

lib/nimble_template/addons/test_env.ex
byhbt: The main purpose of `edit_test_support_case` here is about clean the alias `Ecto.Adapters.SQL.Sandbox`. But on the new version, the `alias` is not used anymore. So I think we can remove the `conn_case` in this case.

They reference directly to the `datacase`:

```
setup tags do            
    SampleApp.DataCase.setup_sandbox(tags)            
                    
    {:ok, conn: Phoenix.ConnTest.build_conn()}            
end
```

Reference: https://www.phoenixdiff.org/?source=1.6.6&target=1.6.11 `test/support/conn_case.ex`

Nihisil (approved):

lib/nimble_template/addons/variants/phoenix/mix_release.ex
andyduong1920: Got it @byhbt

lib/nimble_template/addons/test_env.ex
andyduong1920: Got it @byhbt

andyduong1920 (approved):

lib/nimble_template/addons/variants/phoenix/mix_release.ex
andyduong1920: Got it @byhbt

PR #256, opened 2022-09-07
[Chore] Update credo config to support check private vs public functions order

byhbt:
## What happened 👀

- Update the `.credo.exs` to support the convention `Prefer placing all private functions at the bottom of the module.` on Compass:

image

Link: https://nimblehq.co/compass/development/code-conventions/elixir/#module

## Insight 📝

We need to ignore the check for `ignore: ~w/callback_impl/a` because the generated template Phoenix does not consistently follow this layout, for example:

- https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_web/telemetry.ex#L9
- https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/socket/serializers/v2_json_serializer.ex#L37

Therefore, we ignored the `callback_impl` to keep the `.credo.exs` follow the existing output of the generated Phoenix structure, just support for the `public` and `private` method function order verification.

## Proof Of Work 📹

The tests should pass and the `.credo.exs` output should contain `public_fun private_fun`

image


priv/templates/nimble_template/.credo.exs
```diff
@@ -158,8 +158,8 @@
         #
         {Credo.Check.Readability.StrictModuleLayout,
          [
-           order:
-             ~w/shortdoc moduledoc behaviour use import alias require module_attribute defstruct callback/a
+          order: ~w/shortdoc moduledoc behaviour use import alias require module_attribute defstruct callback public_fun private_fun/a,
+          ignore: ~w/callback_impl/a
```
andyduong1920, line 162: should we add this rule into the `template itself` too @byhbt https://github.com/nimblehq/elixir-templates/blob/develop/.credo.exs#L168

hanam1ni (approved):

Nihisil (approved):

andyduong1920 (approved):
Thanks @byhbt

Issue #253, opened 2022-08-29
Remove the reference to `ExVCR.Mock` in FeatureCase by default


byhbt:
## Issue

Although don't choose to install the `ExVCR.Mock`, the `FeatureCase` still references the `ExVCR.Mock` library. 

![image](https://user-images.githubusercontent.com/948856/187151617-c82e7ada-d1eb-4388-9f99-5bd45e8fe0bb.png)

## Expected

- The `FeatureCase` only includes the `ExVCR.Mock` when the user choose installing the `ExVCR.Mock`.
 
## Steps to reproduce

1. Generate new web variant project, choose no when initializing the addons
Would you like to add the ExVCR addon? [Yn] - [n]

2. Run `mix test`

3. The CLI has this error:

```
==> acme_platform
Compiling 20 files (.ex)
Generated acme_platform app

== Compilation error in file test/acme_platform_web/features/home_page/view_home_page_test.exs ==
** (CompileError) test/acme_platform_web/features/home_page/view_home_page_test.exs:2: module ExVCR.Mock is not loaded and could not be found
    (elixir 1.13.4) expanding macro: Kernel.use/2
    test/acme_platform_web/features/home_page/view_home_page_test.exs:2: AcmePlatformWeb.HomePage.ViewHomePageTest (module)
    (acme_platform 0.1.0) expanding macro: AcmePlatformWeb.FeatureCase.__using__/1
    test/acme_platform_web/features/home_page/view_home_page_test.exs:2: AcmePlatformWeb.HomePage.ViewHomePageTest (module)
    (elixir 1.13.4) expanding macro: Kernel.use/1
```

PR #264, opened 2022-09-19
[gh-253] Do not render the ExVCR in FeatureCase

andyduong1920:
https://github.com/nimblehq/elixir-templates/issues/253

## What happened 👀

Do not render the `use ExVCR.Mock, adapter: ExVCR.Adapter.Hackney` in Feature test if the user hasn't choose the ExVCR addon

## Insight 📝

Introduce the `installed_addons` in the Project where we can store the installed addon.

In the Wallaby addon, add a condition to check if the user chooses the ExVCR addon or not by checking the `installed_addons` contains the `NimbleTemplate.Addons.Phoenix.ExVCR` or not.

## Proof Of Work 📹

The test is passed

### DIFF
--- a/lib/nimble_template/addons/variants/phoenix/web/wallaby.ex
+++ b/lib/nimble_template/addons/variants/phoenix/web/wallaby.ex
@@ -3,6 +3,8 @@ defmodule NimbleTemplate.Addons.Phoenix.Web.Wallaby do
 
   use NimbleTemplate.Addons.Addon
 
+  alias NimbleTemplate.Addons.Phoenix.ExVCR
+
   @impl true
   def do_apply(%Project{} = project, _opts) do
     project
@@ -29,12 +31,17 @@ defmodule NimbleTemplate.Addons.Phoenix.Web.Wallaby do
   end
 
   defp copy_files(
-         %Project{web_module: web_module, base_module: base_module, web_test_path: web_test_path} =
-           project
+         %Project{
+           web_module: web_module,
+           base_module: base_module,
+           web_test_path: web_test_path,
+           installed_addons: installed_addons
+         } = project
        ) do
     binding = [
       web_module: web_module,
-      base_module: base_module
+      base_module: base_module,
+      with_ex_vcr?: ExVCR in installed_addons
     ]
 
     files = [
--- a/lib/nimble_template/addons/addon.ex
+++ b/lib/nimble_template/addons/addon.ex
@@ -16,11 +16,14 @@ defmodule NimbleTemplate.Addons.Addon do
       alias NimbleTemplate.Generator
       alias NimbleTemplate.Hex.Package
       alias NimbleTemplate.Projects.Project
+      alias NimbleTemplate.ProjectHelper
 
       def apply(%Project{} = project, opts \\ %{}) when is_map(opts) do
         Generator.print_log("* applying ", inspect(__MODULE__))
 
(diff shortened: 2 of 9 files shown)


lib/nimble_template/addons/variants/phoenix/web/wallaby.ex
```diff
@@ -29,12 +31,17 @@ defmodule NimbleTemplate.Addons.Phoenix.Web.Wallaby do
...
+           base_module: base_module,
+           web_test_path: web_test_path,
+           installed_addons: installed_addons
+         } = project
        ) do
     binding = [
       web_module: web_module,
-      base_module: base_module
+      base_module: base_module,
+      with_ex_vcr?: ExVCR in installed_addons
```
Nihisil, line 44: Could it be the situation when one module depends on multiple installed addons? So, maybe we can make `installed_addons` a separate module with methods to check enabled addons instead of using them as an array.

What do you think?

andyduong1920: Hi @Nihisil could you clarify your idea a bit are you suggesting to

1/ Change the `installed_addons`'s implementation?
2/ Change the `ExVCR in installed_addons` logic?

Nihisil: I suggested both. We can have something like this

```
defmodule InstalledAddons do
  def add_addon() ...
  def is_ex_vcr_added?() ...
end
```

With that, we will encapsulate logic in one place, and if some addons would require more checks: for example, it may depend on multiple addons and rely on some addons properties.

andyduong1920: and might I know the logic for `is_ex_vcr_added?()` can we still reuse `ExVCR in installed_addons` 🤔

```elixir
defmodule InstalledAddons do
  def add_addon() ...
  
  def is_ex_vcr_added? do
    ExVCR in installed_addons
  end
end
```

Nihisil: Yep, it is correct. My point for these changes is to have all the `is_...` methods in one place instead of using `ExVCR in installed_addons` across the whole code base.

What do you think?

andyduong1920: That is much better for sure, will find sometime to refactor this :D

nvminhtue (approved):

lib/nimble_template/addons/variants/phoenix/web/wallaby.ex
hanam1ni: @andyduong1920 How about we do refactoring after [this RFC](https://github.com/nimblehq/elixir-templates/discussions/261) (https://github.com/nimblehq/elixir-templates/discussions/261) is finalized? so the part to check whether add-ons are added or not can utilize the changes from the RFC.

hanam1ni (approved):

Nihisil (approved):

lib/nimble_template/addons/variants/phoenix/web/wallaby.ex
andyduong1920: Yeah that works as well @hanam1ni cc @Nihisil

byhbt (approved):

PR #276, opened 2022-09-28
[#277] Add Compass Credo check plugin

### LINKED ISSUE #277: Add `compass_credo_plugin` addon
## Why

Integrate [compass_credo_plugin](https://hex.pm/packages/compass_credo_plugin) to the ElixirTemplate.

https://hex.pm/packages/compass_credo_plugin

This library helps enforce [Nimble Elixir coding convention] (https://nimblehq.co/compass/development/code-conventions/elixir/) in the Elixir project.

## Who Benefits?

Developers

hanam1ni:
- Closes #277 

## What happened 👀

Add Compass Credo plugin adding custom checks that comply with our Compass convention.
Repository: https://github.com/nimblehq/compass-credo-plugin

## Proof Of Work 📹

Custom checks are working on generated project

image

### DIFF
--- a/lib/nimble_template/addons/credo.ex
+++ b/lib/nimble_template/addons/credo.ex
@@ -23,9 +23,11 @@ defmodule NimbleTemplate.Addons.Credo do
   end
 
   defp inject_mix_dependency(project) do
-    Generator.inject_mix_dependency(
-      {:credo, latest_package_version(:credo), only: [:dev, :test], runtime: false}
-    )
+    Generator.inject_mix_dependency([
+      {:credo, latest_package_version(:credo), only: [:dev, :test], runtime: false},
+      {:compass_credo_plugin, latest_package_version(:compass_credo_plugin),
+       only: [:dev, :test], runtime: false}
+    ])
 
     project
   end
--- a/lib/nimble_template/helpers/credo.ex
+++ b/lib/nimble_template/helpers/credo.ex
@@ -1,0 +1,10 @@
+defmodule NimbleTemplate.CredoHelper do
+  alias NimbleTemplate.Generator
+
+  @spec disable_rule(String.t(), String.t()) :: :ok | {:error, :failed_to_read_file}
+  def disable_rule(file_path, rule) do
+    Generator.prepend_content(file_path, """
+    # credo:disable-for-this-file #{rule}
+    """)
+  end
+end
--- a/lib/nimble_template/templates/template.ex
+++ b/lib/nimble_template/templates/template.ex
@@ -4,6 +4,7 @@ defmodule NimbleTemplate.Templates.Template do
   import NimbleTemplate.DependencyHelper
 
   alias NimbleTemplate.Addons.ExUnit
+  alias NimbleTemplate.CredoHelper
   alias NimbleTemplate.Projects.Project
   alias NimbleTemplate.Templates.Mix.Template, as: MixTemplate
   alias NimbleTemplate.Templates.Phoenix.Template, as: PhoenixTemplate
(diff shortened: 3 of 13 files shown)


lib/nimble_template/helpers/credo.ex
```diff
@@ -0,0 +1,10 @@
+defmodule NimbleTemplate.CredoHelper do
+  alias NimbleTemplate.Generator
+
+  @spec disable_rule(String.t(), String.t()) :: :ok | {:error, :failed_to_read_file}
+  def disable_rule(file_path, rule) do
```
nvminhtue, line 5: do we need to check the existence of the input `rule` @hanam1ni 🤔

hanam1ni: @nvminhtue IMO, we don't need to check because if a `rule` is not given, only single param is passed to this function and it will raise `FunctionClauseError`. And if an empty string is passed here, the function can work properly and doesn't break the app.

lib/nimble_template/addons/credo.ex
```diff
@@ -23,9 +23,11 @@ defmodule NimbleTemplate.Addons.Credo do
   end
 
   defp inject_mix_dependency(project) do
-    Generator.inject_mix_dependency(
-      {:credo, latest_package_version(:credo), only: [:dev, :test], runtime: false}
-    )
+    Generator.inject_mix_dependency([
+      {:credo, latest_package_version(:credo), only: [:dev, :test], runtime: false},
+      {:compass_credo_plugin, latest_package_version(:compass_credo_plugin),
```
andyduong1920, line 28: Should we add this into the `credo` test @hanam1ni

lib/nimble_template/templates/template.ex
```diff
@@ -52,6 +56,37 @@ defmodule NimbleTemplate.Templates.Template do
...
+      [
+        "lib/#{Macro.underscore(base_module)}.ex",
+        "#{web_path}/controllers/page_controller.ex",
+        "#{web_path}/telemetry.ex",
+        "#{web_path}/views/error_view.ex",
+        "#{web_path}/views/error_view.ex",
+        "#{base_path}/release_tasks.ex"
+      ],
+      fn file_path ->
+        CredoHelper.disable_rule(file_path, "CompassCredoPlugin.Check.DoSingleExpression")
```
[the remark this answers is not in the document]
hanam1ni: IMO, we should disable the rule for the entire file instead of fixing these file content because these files are generated from the Phoenix generator, and this template can be used with different Phoenix generator versions depending on the user's machine. The different generator versions can result in different file content, so we should disable the rule for the file.

What do you think? 💭

lib/nimble_template/addons/credo.ex
hanam1ni: Added test in [297a251](https://github.com/nimblehq/elixir-templates/pull/276/commits/297a251f9e02e3b223b4e799f2c0a9bca52294b5)

test/nimble_template/addons/credo_test.exs
```diff
@@ -28,6 +28,7 @@ defmodule NimbleTemplate.Addons.CredoTest do
                    defp deps do
                      [
                        {:credo, "~> 1.4", [only: [:dev, :test], runtime: false]},
+                       {:compass_credo_plugin, "~> 1.0.0", [only: [:dev, :test], runtime: false]},
```
andyduong1920, line 31: Should we avoid the call to Hex.pm to get the `compass_credo_plugin` version by using the `@describetag mock_latest_package_versions: [{:compass_credo_plugin, "1.0.0"}]` @hanam1ni

andyduong1920: I'm opening a PR for this issue as well FYI @hanam1ni https://github.com/nimblehq/elixir-templates/pull/284

lib/nimble_template/templates/template.ex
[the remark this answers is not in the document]
andyduong1920: I'm not sure @hanam1ni as we would like to use the template to generate the project and we should reduce the `manual changes` as much as we can on the `output project` so with the `disable credo rule` like this, later on the people will need to remove it and adjust the file manually 🤔

test/nimble_template/addons/credo_test.exs
hanam1ni: @andyduong1920 Good catch 👍 I updated in [a7416d4](https://github.com/nimblehq/elixir-templates/pull/276/commits/a7416d412209f50c9167b42fb14970a0ec7a4847)

Nihisil (approved):

lib/nimble_template/templates/template.ex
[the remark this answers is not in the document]
hanam1ni: @andyduong1920 Some files that are generated with Phoenix generator are not comply with our Compass convention, such as some of the functions in `telemetry.ex` or `release_tasks.ex` does not follow our convention [here](https://nimblehq.co/compass/development/code-conventions/elixir/#convention-20). 
```elixir
# telemetry.ex
defmodule ProjectWeb.Telemetry do
  def start_link(arg) do
    Supervisor.start_link(__MODULE__, arg, name: __MODULE__)
  end

  # ...
end
```
So we need to disable rules in some generated files, which we don't have control over them, to make prevent Credo raise any warning from the fresh generated project and also to pass the CI.

andyduong1920 (approved):

longnd (approved):

PR #284, opened 2022-11-09
Do not allow the Test ENV call Hex.pm directly

andyduong1920:
## What happened 👀

Currently, if the developer forgets to mock the external library, our test will send a real request to Hex.pm to get the latest version this brings two issues

- The unit test could fail in the future if the 3rd library has a new version on Hex.pm
- Waste 1 API call to external service

Related comment https://github.com/nimblehq/elixir-templates/pull/276/files#r1017442442

## Insight 📝

Using different resources per each environment.
- Dev, Prod: Package
- Test: PackageMock

On the Test ENV, if we missed the `@describetag mock_latest_package_versions: [{xx, yy}]` our test won't pass.

### Note

There is a workaround I'm not sure why that application environment is nil when we apply the template.

image
 

## Proof Of Work 📹

1/ The test is pass
2/ Tested on my local

image

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -2,8 +2,7 @@ import Config
 
 config :phoenix, :json_library, Jason
 
-config :nimble_template,
-  hex_package_resource: NimbleTemplate.Hex.Package
+config :nimble_template, hex_package_resource: NimbleTemplate.Hex.Package
 
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,3 +1,10 @@
 import Config
 
 config :phoenix, :json_library, Jason
+
+config :nimble_template,
+  hex_package_resource: NimbleTemplate.Hex.Package
+
+# Import environment specific config. This must remain at the bottom
+# of this file so it overrides the configuration defined above.
+import_config "#{config_env()}.exs"
--- a/config/config.exs
+++ b/config/config.exs
@@ -4,4 +4,6 @@ config :phoenix, :json_library, Jason
 
 config :nimble_template, hex_package_resource: NimbleTemplate.Hex.Package
 
-import_config "#{config_env()}.exs"
+if File.exists?("config/#{config_env()}.exs") do
+  import_config "#{config_env()}.exs"
+end
--- a/config/config.exs
+++ b/config/config.exs
@@ -4,6 +4,4 @@ config :phoenix, :json_library, Jason
 
 config :nimble_template, hex_package_resource: NimbleTemplate.Hex.Package
 
-# Import environment specific config. This must remain at the bottom
-# of this file so it overrides the configuration defined above.
 import_config "#{config_env()}.exs"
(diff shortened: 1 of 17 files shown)


lib/nimble_template/hex/package_mock.ex
```diff
@@ -0,0 +1,14 @@
+defmodule NimbleTemplate.Hex.PackageMock do
+  @moduledoc false
+
+  @behaviour NimbleTemplate.Hex.PackageBehaviour
+
+  alias NimbleTemplate.Exception.MockHexPackageRequiredException
+
+  def get_latest_version(package) do
```
longnd, line 8: is it worth leaving a comment here, explaining the reason that we need to mock this function - to prevent the call to get the latest version if the test forgets to mock external library?

longnd (approved):

config/config.exs
```diff
@@ -1,3 +1,7 @@
 import Config
 
 config :phoenix, :json_library, Jason
+
+config :nimble_template, hex_package_resource: NimbleTemplate.Hex.Package
+
+import_config "#{config_env()}.exs"
```
hanam1ni, line 7: How about adding a check if the config exists or not? so we don't need to create a blank configuration file for dev and prod.
```suggestion
if File.exists?("config/#{config_env()}.exs") do
  import_config "#{config_env()}.exs"
end
```

lib/nimble_template/exception/mock_hex_package_required.ex
```diff
@@ -0,0 +1,3 @@
+defmodule NimbleTemplate.Exception.MockHexPackageRequiredException do
```
hanam1ni, line 1: Should we place this exception under `NimbleTemplate.Hex` namespace?

```suggestion
defmodule NimbleTemplate.Hex.Exception.MockHexPackageRequiredException do
```

lib/nimble_template/hex/package_mock.ex
```diff
@@ -0,0 +1,14 @@
...
+    raise(MockHexPackageRequiredException,
+      message:
+        "Mock Hex Package with `@describetag mock_latest_package_versions: [{#{package}, :version}]` required"
```
hanam1ni, line 11: How about?
```suggestion
        "Requires to mock Hex package with `@describetag mock_latest_package_versions: [{#{package}, :version}]`"
```

lib/nimble_template/exception/mock_hex_package_required.ex
andyduong1920: Based on our structure on the compass, we prefer to keep the exception outside the context @hanam1ni 

image

lib/nimble_template/hex/package_mock.ex
andyduong1920: added in https://github.com/nimblehq/elixir-templates/pull/284/commits/7a0aff5ce47ddb7c7245d660788fe44a6c8ffe76 @longnd

lib/nimble_template/hex/package_mock.ex
andyduong1920: updated in https://github.com/nimblehq/elixir-templates/pull/284/commits/0a89e641f577bdf2e4e0b1545e3dea1f7567369f :)

config/config.exs
andyduong1920: updated in https://github.com/nimblehq/elixir-templates/pull/284/commits/0a89e641f577bdf2e4e0b1545e3dea1f7567369f

hanam1ni (approved):
LGTM ✨

Nihisil (approved):

Issue #249, opened 2022-08-22
Fix: Infinite loop in the `Upgrade stack` task


andyduong1920:
## Issue

It seems like the Replace content has an infinite loop issue

https://github.com/nimblehq/elixir-templates/actions/runs/2900713463
 
## Expected

The Upgrade stack task should work well
 
## Steps to reproduce

1. Trigger the https://github.com/nimblehq/elixir-templates/actions/workflows/upgrade_stack.yml
2. Enter the following versions: `--elixir 1.13.4 --erlang 25.0.4 --node 16.17.0 --alpine 3.16.1`
3. Run the Workflow

PR #285, opened 2022-11-09
[gh-249] Fix infinity loop on Upgrade stack

andyduong1920:
https://github.com/nimblehq/elixir-templates/issues/249

## What happened 👀

Solved https://github.com/nimblehq/elixir-templates/issues/249

## Insight 📝

It's because the `anchor is equal to the content`, so that method keep loop forever.

In our case, it's because we have the `-otp-25` in existing Elixir version, and when we upgrade to another `erlang 25` version, we try to `find and replace the -otp-25 by -otp-25`

image

## Proof Of Work 📹

1. Tested on my local
2. Trigger the Workflow https://github.com/nimblehq/elixir-templates/actions/runs/3426992297/jobs/5709447990
3. We have the PR created by that workflow https://github.com/nimblehq/elixir-templates/pull/286

### DIFF
--- a/lib/nimble_template/helpers/generator.ex
+++ b/lib/nimble_template/helpers/generator.ex
@@ -24,7 +24,12 @@ defmodule NimbleTemplate.Generator do
 
   def rename_file(old_path, new_path), do: File.rename(old_path, new_path)
 
-  def replace_content(file_path, anchor, content, raise_exception \\ true) do
+  def replace_content(file_path, anchor, content, raise_exception \\ true)
+
+  def replace_content(_file_path, anchor, content, _raise_exception) when anchor == content,
+    do: :ok
+
+  def replace_content(file_path, anchor, content, raise_exception) do
     file = Path.join([file_path])
 
     file_content =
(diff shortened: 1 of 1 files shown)


lib/nimble_template/helpers/generator.ex
```diff
@@ -24,7 +24,12 @@ defmodule NimbleTemplate.Generator do
 
   def rename_file(old_path, new_path), do: File.rename(old_path, new_path)
 
-  def replace_content(file_path, anchor, content, raise_exception \\ true) do
+  def replace_content(file_path, anchor, content, raise_exception \\ true)
+
+  def replace_content(_file_path, anchor, content, _raise_exception) when anchor == content,
```
liamstevens111, line 29:
```suggestion
  def replace_content(_file_path, anchor, anchor, _raise_exception),
```
I think this also works but yours I think is more clear what it's doing? 👁️

liamstevens111 (approved):
Comment/Question but nice work 👍

lib/nimble_template/helpers/generator.ex
andyduong1920: that isn't clear enough I think @liamstevens111 and I'm not sure it work as the same 😂 so let go with `when anchor == content`

liamstevens111: Yeah should work the same but I agree, probably better to be explicit here @andyduong1920

byhbt (approved):

Issue #307, opened 2022-12-27
Fix invalid disable Credo rule on not existing file


byhbt:
## Issue

When generating a new API project with the command `mix phx.new sample_project --no-html`, we got the message:

<img width="1058" alt="Screen Shot 2022-12-27 at 14 39 08" src="https://user-images.githubusercontent.com/948856/209629925-cd39243f-c843-4e57-b466-2480d19b69af.png">

Because the `--no-html` params won't create the `page_controller.ex` so the patching disable single expression failed. 

To prevent this case, we need to check if the file which is patching disable credo rule has existed in the existing codebase or not.

## Expected

The message should not be shown.
 
## Steps to reproduce

1. Create a new project with `mix phx.new sample_project --no-html`
1. Then apply `mix nimble_template.gen --api`

liamstevens111:
Good catch, I thought I manually tested with `api` project too, guess I only did `mix` and `phoenix`, sorry about that.

PR #308, opened 2022-12-28
[gh-307] Fix the invalid disable Credo rule on API variant

byhbt:
Resolves https://github.com/nimblehq/elixir-templates/issues/307

## What happened 👀

- [x] Extract the disable Credo rule into the credo helper.
- [x] Add checking `file_exist?/1` in case a new project is created with `--no-html`, the `page_controller` does not exist there.
- [x] Rename  fetch_and_install_* => `install_*` to make it cleaner.
- [x] Refactor the `template.ex`: using `alias` to reference to the helper methods instead of `import`

## Insight 📝

n/a

## Proof Of Work 📹

The test should pass.

### DIFF
--- a/lib/nimble_template/helpers/credo.ex
+++ b/lib/nimble_template/helpers/credo.ex
@@ -7,9 +7,11 @@ defmodule NimbleTemplate.CredoHelper do
 
   @spec disable_rule(String.t(), String.t()) :: :ok | {:error, :failed_to_read_file}
   def disable_rule(file_path, rule) do
-    Generator.prepend_content(file_path, """
-    # credo:disable-for-this-file #{rule}
-    """)
+    if File.exists?(file_path) do
+      Generator.prepend_content(file_path, """
+      # credo:disable-for-this-file #{rule}
+      """)
+    end
   end
 
   @spec disable_do_single_expression_rule(list()) :: :ok | {:error, :failed_to_read_file}
@@ -38,38 +40,39 @@ defmodule NimbleTemplate.CredoHelper do
       ) do
     suppress_credo_warnings_for_base_project(project)
 
-    single_expression_files = [
+    files_contain_single_expression = [
       "#{web_path}/controllers/page_controller.ex",
       "#{web_path}/telemetry.ex",
       "#{web_path}/views/error_view.ex",
       "#{base_path}/release_tasks.ex"
     ]
 
-    disable_do_single_expression_rule(single_expression_files)
+    disable_do_single_expression_rule(files_contain_single_expression)
   end
 
   def suppress_credo_warnings_for_phoenix_api_project(
         %Project{
-          web_test_path: web_test_path,
           base_path: base_path,
-          web_path: web_path
+          web_path: web_path,
+          web_test_path: web_test_path
         } = project
       ) do
     suppress_credo_warnings_for_base_project(project)
 
-    single_module_files = [
+    files_contain_multiple_module = [
       "#{web_test_path}/views/api/error_view_test.exs",
       "#{web_test_path}/params/params_validator_test.exs"
     ]
 
-    disable_single_module_file_rule(single_module_files)
+    disable_single_module_file_rule(files_contain_multiple_module)
 
-    single_expression_files = [
+    files_contain_single_expression = [
+      "#{web_path}/controllers/page_controller.ex",
       "#{base_path}/release_tasks.ex",
       "#{web_path}/telemetry.ex",
       "#{web_path}/views/error_view.ex"
     ]
 
-    disable_do_single_expression_rule(single_expression_files)
+    disable_do_single_expression_rule(files_contain_single_expression)
   end
 end
--- a/lib/nimble_template/helpers/credo.ex
+++ b/lib/nimble_template/helpers/credo.ex
@@ -32,47 +32,44 @@ defmodule NimbleTemplate.CredoHelper do
     disable_do_single_expression_rule(["lib/#{Macro.underscore(base_module)}.ex"])
   end
 
-  def suppress_credo_warnings_for_phoenix_project(
-        %Project{
-          base_path: base_path,
-          web_path: web_path
-        } = project
-      ) do
+  def suppress_credo_warnings_for_phoenix_project(project) do
     suppress_credo_warnings_for_base_project(project)
 
-    files_contain_single_expression = [
-      "#{web_path}/controllers/page_controller.ex",
-      "#{web_path}/telemetry.ex",
-      "#{web_path}/views/error_view.ex",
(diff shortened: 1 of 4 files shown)


lib/nimble_template/helpers/credo.ex
```diff
@@ -1,10 +1,75 @@
...
+      disable_rule(file_path, @do_single_expression_rule_name)
+    end)
+  end
+
+  @spec disable_single_module_file_rule(list()) :: :ok | {:error, :failed_to_read_file}
+  def disable_single_module_file_rule(file_paths) do
+    Enum.each(file_paths, fn file_path ->
+      disable_rule(file_path, @single_module_file_rule_name)
+    end)
+  end
```
liamstevens111, line 29: As `disable_rule`, `disable_do_single_expression_rule` and `disable_single_module_file_rule` functions are not called from outside of this module and only from functions below shall we move it to towards the end and make them private? Or do you think there's plans for it to be called from outside the module in the future?

I know this existed on disable_rule already but do we need to add `@spec` to these? I think we don't use this consistently throughout the template project or is there only specific cases where we use it?

lib/nimble_template/helpers/credo.ex
```diff
@@ -1,10 +1,75 @@
...
+
+  def suppress_credo_warnings_for_base_project(%Project{base_module: base_module}) do
+    disable_do_single_expression_rule(["lib/#{Macro.underscore(base_module)}.ex"])
+  end
```
liamstevens111, line 33: Shall we move this to a `, do:` instead of `do...end` as it only contains one line? This is the convention that the check you're disabling will do in real projects :trollface:

liamstevens111 (commented):
Nice find on the bug occurring when `--no-html` is used to generate an API project as I only tested on the example creations given in the README for the template, ie

```elixir
# New Phoenix project
mix phx.new awesome_project

# New Mix project
mix new awesome_project
```

👍 👍 

Few comments on pre-existing code and on the credo helper 👌

lib/nimble_template/helpers/credo.ex
```diff
@@ -1,10 +1,75 @@
...
+      """)
+    end
+  end
+
+  @spec disable_do_single_expression_rule(list()) :: :ok | {:error, :failed_to_read_file}
+  def disable_do_single_expression_rule(file_paths) do
+    Enum.each(file_paths, fn file_path ->
+      disable_rule(file_path, @do_single_expression_rule_name)
+    end)
+  end
```
liamstevens111, line 22: Instead of having a function for each rule that loops file_paths and calls `disable_rule` what about a more generic function `disable_rules` with an `s` that you pass the rule which calls `disable_rule` on each. Then we can just use:

`disable_rules(file_paths, @do_single_expression_rule_name)` 
and
`disable_rules(file_paths, @single_module_file_rule_name)` 

```elixir
project
    |> get_files_contain_multiple_modules()
    |> disable_single_module_file_rule()
```

vs

```elixir
project
    |> get_files_contain_multiple_modules()
    |> disable_rules(@do_single_expression_rule_name)
```
or something similiar

lib/nimble_template/templates/template.ex
```diff
@@ -31,82 +30,35 @@ defmodule NimbleTemplate.Templates.Template do
...
   end
 
-  defp fetch_and_install_elixir_dependencies() do
+  defp install_elixir_dependencies() do
     Mix.shell().cmd("MIX_ENV=dev mix do deps.get, deps.compile")
     Mix.shell().cmd("MIX_ENV=test mix do deps.get, deps.compile")
   end
 
-  defp fetch_and_install_node_dependencies() do
+  defp install_node_dependencies() do
```
nvminhtue, line 58: It should reformat to `,do: ...` block and the `format_codebase` function. What do you think?

nvminhtue (approved):

lib/nimble_template/templates/template.ex
```diff
@@ -1,10 +1,9 @@
 defmodule NimbleTemplate.Templates.Template do
   @moduledoc false
 
-  import NimbleTemplate.DependencyHelper
+  import NimbleTemplate.{CredoHelper, DependencyHelper}
 
   alias NimbleTemplate.Addons.ExUnit
-  alias NimbleTemplate.CredoHelper
```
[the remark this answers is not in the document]
byhbt: I just do a quick research to ensure we use the correct approach. In this case, we should use `alias`
- Readability: the developer have better understand about where the methods come from. "Explicit is better than implicit". Phoenix 1.4 also refactor the same for the Routes module: https://gist.github.com/chrismccord/bb1f8b136f5a9e4abc0bfc07b832257e#add-a-routes-alias-and-update-your-router-calls

- Second is about the coupling between module: https://stackoverflow.com/questions/70107659/in-elixir-why-is-alias-preferred-over-import-for-importing-the-modules

So I think `alias` is better approach for this case.

-

lib/nimble_template/helpers/credo.ex
byhbt: Good suggestion, refactored in https://github.com/nimblehq/elixir-templates/pull/308/commits/ce27c66a70f9c83889b6fe91e171cd5a03d244e4

lib/nimble_template/helpers/credo.ex
byhbt: Nice suggestion, refactored in https://github.com/nimblehq/elixir-templates/pull/308/commits/ce27c66a70f9c83889b6fe91e171cd5a03d244e4

lib/nimble_template/templates/template.ex
byhbt: Nice catch, updated in https://github.com/nimblehq/elixir-templates/pull/308/commits/ce27c66a70f9c83889b6fe91e171cd5a03d244e4

lib/nimble_template/helpers/credo.ex
byhbt: Updated in https://github.com/nimblehq/elixir-templates/pull/308/commits/ce27c66a70f9c83889b6fe91e171cd5a03d244e4

lib/nimble_template/templates/template.ex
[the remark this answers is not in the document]
byhbt: cc @Nihisil @hanam1ni @andyduong1920 @nvminhtue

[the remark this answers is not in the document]
nvminhtue: They are all good to me either way. IMO, `alias` sounds a bit more appropriate to distinguish with several private functions declared in this module.

Nihisil (approved):

liamstevens111 (approved):

PR #323, opened 2023-03-17
Convert from ErrorView to new ErrorJSON view

liamstevens111:
https://github.com/nimblehq/elixir-templates/issues/??

## What happened 👀

Completed:
- ErrorHelper doesn't exist on Phoenix `1.7` as I think moved to Component along with translations. We had the default file with the addition of the function `status_code_from_template`. I have moved this  function to be inside ErrorJSON
- The previous `ErrorView.template_not_found` is now `ErrorJSON.render`

Outstanding Issues:
- Edited ViewCase (Route.Helpers > VerifiedRoutes). It seems we do not even generate this file in output though?
- `translate_error` that use to be in ErrorHelper has been moved to CoreComponents in 1.7 for use with input component. I tried to import this but module not found so redefined the functions in ErrorJSON

## Insight 📝

`N/A`

## Proof Of Work 📹

New project with test API for response:

Screenshot 2023-03-21 at 08 53 58

The tests we had before are passing for the response


priv/templates/nimble_template/lib/otp_app_web/controllers/error_json.ex.eex
```diff
@@ -39,17 +71,13 @@ defmodule <%= web_module %>.Api.ErrorView do
...
-          message: message
-        }
-      ]
-    }
-  end
-
   defp translate_errors(changeset), do: Changeset.traverse_errors(changeset, &translate_error/1)
+
+  defp translate_error({msg, opts}) do
+    if count = opts[:count] do
```
nvminhtue, line 77: isn't it?
```suggestion
    if count == opts[:count] do
```

liamstevens111: It assigns count at the same time for use in the positive case so it's correct.

See original implementation here:
https://github.com/nimblehq/elixir-templates/blob/25fd192d8810348e9a0ba2579b4616f2d2952107/priv/templates/nimble_template/lib/otp_app_web/views/error_helpers.ex.eex#L43-L48

This originates from Phoenix as well

Good observation though 👍

olivierobert (approved):

priv/templates/nimble_template/lib/otp_app_web/controllers/error_json.ex.eex
andyduong1920: shall we mark this one as `Resolved` @nvminhtue

byhbt (approved):

priv/templates/nimble_template/test/support/view_case.ex.eex
```diff
@@ -3,14 +3,14 @@ defmodule <%= web_module %>.ViewCase do
 
   using do
     quote do
+      use <%= web_module %>, :verified_routes
+      # TODO: use Phoenix.Controller for previous View.render?
```
andyduong1920: what does this TODO mean @liamstevens111 will we have a plan to fix this 🤔 

The new generated project will have this TODO, not sure how will we handle this

andyduong1920 (approved):

PR #324, opened 2023-03-28
Replace the cancel previous run by the concurrency

byhbt:
## What happened 👀

- Replace the step `Cancel Previous Runs` in all the workflow. Replace it by the `concurrency` group
- Add new team member to the CODEOWNER

## Insight 📝

As the suggestion from the creator of `Cancel Previous Runs`:

image

Reference: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

## Proof Of Work 📹

image

### DIFF
--- a/priv/templates/nimble_template/.github/workflows/deploy_heroku.yml
+++ b/priv/templates/nimble_template/.github/workflows/deploy_heroku.yml
@@ -17,7 +17,7 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' || github.ref != 'refs/heads/develop' }}
 
 jobs:
   deploy:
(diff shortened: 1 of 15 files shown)


priv/templates/nimble_template/.github/workflows/deploy_heroku.yml
```diff
@@ -15,17 +15,16 @@ env:
   HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
   HEROKU_APP_NAME: ${{ secrets.HEROKU_APP_NAME }}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```
andyduong1920: shall we cancel it on the `develop` branch as well 🤔 because we deploy to Staging from the develop

olivierobert (commented):
@byhbt let's complete this PR as there is not much left to do, no?

byhbt:
> @byhbt let's complete this PR as there is not much left to do, no?

yup, i was checking about the Andy suggestion 😅 , let me update it and make it ready for review today or tomorrow.

